PR #41, opened 2015-02-20
added map_type 8 for angular mapping aka Debevec mirrorball HDR 

Lefinnois:
Hi,
Here is some code to add a new mapping type.
This is "map_type 7" in MegaPov (7 is allready used, so i switched from 7 to 8). It can be used for mapping angular maps to a sphere.
This is the kind of mapping Paul Debevec use for HDR images made from pictures taken of reflective ball. With this patch i can render my old pov files made for MegaPov, just by changing 7 to 8, and "ambient n" to "emission n" in finish.


c-lipka:
Thanks for sharing! This is one of the things that have been on my ToDo list but I just couldn't get round to it. I'll have a closer look at the code and do a pull in the next couple of days.

c-lipka:
FYI, I'll revert the Angular Map back to map type 7. While POV-Ray did indeed define a constant for that map type, it didn't map anything to it.

Lefinnois:
You are talking about OLD_MAP. I saw it, yes. But I just did not want to remove an existing definition. Type 7 would be great.

PR #56, opened 2016-03-12
Update vector.h

wfpokorny:
Use of abs() instead of fabs() in vector functions recently created for the new user defined camera caused problems for Ubuntu users. Edit changes to fabs(). See: http://news.povray.org/povray.text.scene-files/message/%3C56e36723%241%40news.povray.org%3E/#%3C56e36723%241%40news.povray.org%3E


c-lipka:
Thanks for analyzing and fixing.

Issue #60, opened 2016-05-02
Not the right behaviour for Bezier splines 


Lanuhum:
http://news.povray.org/povray.binaries.images/message/%3Cweb.5727406d27a8e5387a3e03fe0%40news.povray.org%3E/#%3Cweb.5727406d27a8e5387a3e03fe0%40news.povray.org%3E

File '/tmp/Empty.pov' line 17: Parse Error: Incorrect point in lathe.
Fatal error in parser: Cannot parse input.
Render failed

#version 3.7;

#include "functions.inc"
global_settings {
    assumed_gamma 1.000000
    max_trace_level 3
    charset utf8
}
sky_sphere {
    pigment {rgb<0.050, 0.050, 0.050>}
}

#declare Default_texture = texture{pigment {rgb 0.8}}

#declare data_Lathe_shape_ob =
lathe { bezier_spline 8,
    <0,1><0.25,0.75><-0.07594,0.5><0.1875,0.1875>
    <0.1875,0.1875><0.3929,-0.05616><0,-0.5><0,-1>
}
object {data_Lathe_shape_ob
    texture {Default_texture}
    matrix <1.000000, 0.000000, 0.000000,  0.000000, -0.000000, -1.000000,  0.000000, 1.000000, -0.000000,  0.000000, 0.000000, 0.000000>
}

light_source {
    <5.07,5.58,4.28>
    color rgb<1, 1, 1>
    fade_distance 25.0000000000
    fade_power 1
}
camera {
    perspective
    location  <0,0,0>
    look_at  <0,0,-1>
    right <-1.7777777777777777, 0, 0>
    up <0, 1, 0>
    angle  49.134343
    rotate  <-27.098163, 46.688390, -0.903519>
    translate <7.481132, 5.343666, 6.507640>
}

wfpokorny:
Attaching a large set of spline related test cases used in part to evaluate a code update to permit negative control points in a bezier spline lathe. 
[lathePrismSplineIssuesTestCases.tar.gz](https://github.com/POV-Ray/povray/files/264502/lathePrismSplineIssuesTestCases.tar.gz)

c-lipka:
I think this should be fixed with commit 2ec4a4e / pull request #62.

PR #62, opened 2016-05-14
Lathe parser updates related to github issue #60

wfpokorny:
In summary, current on curve point checking for positive x is correct only for the cubic spline, wrong for the bezier control points and both its ends, and wrong for one or both ends of the quadratic and linear splines, respectively. The new checking forces all on curve, specified points, to be positive in x. Know that while not too likely, somebody, somewhere, no doubt specified point(s) which worked with the previous code which will now need adjustment at the first or last point.


source/parser/parser.cpp
```diff
@@ -2911,7 +2911,7 @@ ObjectPtr Parser::Parse_Julia_Fractal ()
 
 ObjectPtr Parser::Parse_Lathe()
 {
-    int i;
+    int i, AlreadyWarned;
```
c-lipka: Can you make this a bool, and adjust the test and modification of this variable accordingly? I don't think we need to count how many times we have issued a warning.

wfpokorny:
Hi Christoph, I've made the update. Wondered about updating an active pull request, but git / github made it easy. 

I saw too your requests for a documentation update and for a separate prism->bezier_spline->normal issue to be opened. I am working on a larger test case now for the new prism issue. 

Where do I make / suggest these documentation updates? Is this something that can be done through the wiki page itself? Or still the newsgroup & somebody else does the actual update? Thanks.

c-lipka:
Once you have signed in to the Wiki, you should be able to make changes to the documentation (though some pages have been locked; if that is the case for the portions you intend to provide changes for, please add the suggested new text to the discussion page).

c-lipka:
Whoops, wrong button.

Issue #63, opened 2016-05-18
Mar 05, 2016 commit appears to have broken AA.  Ubuntu 14.04.


wfpokorny:
Something in the changes for the commit c89113156da57f9bd4261bfa24342a461e4521a2

|| Date:   Sat Mar 5 15:31:55 2016 +0100
|| 
||     Fixed two mesh camera bugs and other camera flaws.
||     
||     - Fixed FS#254 (mesh_camera type 0 repeats first pixel line).
||     - Fixed FS#261 (mesh_camera type 3 output image is offset by 0.5 pixels).
||     - Fixed mesh_camera-related flaws in the parser.
||     - Fixed a family of bugs that could cause problems when using focal blur wit
||     - A bit of refactoring (changed `TracePixel::operator()` semantics).

appears to have broken AA. See attached png files for this commit and the last code commit just prior.

Ubuntu 14.04

Comand used: povray +w1200 +h800 +a0.3 +am2 +r3 master371.pov
[master371.pov.txt](https://github.com/POV-Ray/povray/files/270920/master371.pov.txt)
![master371_10465bbc64bc63a0284ed011b7009b6209568e40_ok](https://cloud.githubusercontent.com/assets/15925553/15366239/66f5ed88-1cf1-11e6-96ee-6023466039b5.png)
![master371_c89113156da57f9bd4261bfa24342a461e4521a2_bad](https://cloud.githubusercontent.com/assets/15925553/15366240/66f6cfe6-1cf1-11e6-99ac-cfa5ddc61b3e.png)

c-lipka:
Please check the latest master; it should fix the issue.

wfpokorny:
OK. You can then ignore my pull request which duplicates your update less the version.

wfpokorny:
Merged in your committed code and, post local compile, looks good. Closing.

PR #64, opened 2016-05-18
Fix for AA issue #63 introduced with commit c891131 in March

wfpokorny:
Looks like we missed one change in the parameters to trace back in March.


c-lipka:
Noticed your pull request too late -- sorry, it's already fixed :)

Issue #66, opened 2016-06-04
Lack of 32 bit df3 write capability from the SDL.


wfpokorny:
I have updates in hand locally will shortly be generating a pull request to add 32bit, unsigned, big endian format write capability to the SDL. The 8 bit and 16 bit methods exist already via the ARRAYS_WriteDF3() macro in arrays.inc.

wfpokorny:
Closing as portable methods apparently not currently doable.

PR #67, opened 2016-06-04
Updates to address the lack of 32bit df3 write capability. Issue #66

wfpokorny:
Note, I have too in the works an update to the Density File Pattern reference section which depend upon the SDL capability added in this pull request. To review this update - showing as a side effect my testing - see:   http://wiki.povray.org/content/User:Wfpokorny


wfpokorny:
OK.  Sorry to burn your time.

PR #68, opened 2016-06-06
Pull from current master into my version of povray

wfpokorny:
Pull from current master into my version of povray


wfpokorny:
..... Dang. Started a pull in the wrong direction.

Issue #36, opened 2014-11-20
Bad naming in Insert Menu / 10 - Ready made scenes


gryken2014:
In "10 - Ready made scenes" directory, for example :

10 - Basic Scene 01 - Checkered plane.txt
10 - Basic scene 01 - Checkered plane.jpg

With text files, there is "Scene".
With jpg files, there is "scene".

It is harmless under Windows (non case sensitive file-system)
But, it can be under unix-like (case sensitive file-system)

c-lipka:
These files reside in the "distribution/platform-specific/windows" directory, so I don't see much of a problem there. Leaving the issue open though, so that it won't be entirely forgotten.

PR #73, opened 2016-07-10
Rename scene to Scene - issue #36

gryken2014:
Files renaming (issue #36 )


c-lipka:
Thanks for your submission. Unfortunately it fails one of the checks, probably because your changes are based on an all-too-outdated version. You can probably fix the issue by merging the up-to-date master into your branch `hotfix_bad_names_in_insert_menu`.

gryken2014:
Hello

I have just created a new one, to replace this one.

How can i delete it ?

Thank you.

c-lipka:
When a pull request needs some fixes, it is generally not necessary to close it and re-open another one: Pull requests automatically track any changes you make to the corresponding branch in your repository. Also, there is no need to be afraid of submitting pull requests with a convoluted history: We have the option to "squash and merge" pull requests into a single commit.

As for deleting a pull request, I don't think that can be accomplished in a reasonable manner. To my knowledge, closing a pull request (which you have already done) is as good as it gets.

PR #76, opened 2016-07-14
fix filenames / issue #36

gryken2014:
17 fix in file-names in insert menu :
(only case changes)

Directory : 
"povray/distribution/platform-specific/windows/Insert Menu/10 - Ready made scenes"

Cf issue #36


c-lipka:
Thanks for your contribution!

PR #79, opened 2016-08-18
Feature/new density pattern interpolations

wfpokorny:
For detail see the initial documentation created for this request at: 

http://wiki.povray.org/content/User:Wfpokorny/DensityFile

I've also tried to flush out the df3 support with several new macros currently in another include file and part of the documentation. I think these macros should probably end up in arrays.inc if adopted.

I'll post a couple images to the p.b.i to see if others there will review. I also have a fairly large set of test cases from which about an 18MB review web page is created, but given the size, I'll not post it unless prompted. I'm not all that confident in my c++ skills so any code review appreciated. 

Oh, I picked up one TODO from the code in the ContinuousPattern::Evaluate thinking in working with the density_file pattern I was positioned to look at it. Moved from the 'magic value' to code's EPSILON value.

### DIFF
--- a/source/core/material/pattern.cpp
+++ b/source/core/material/pattern.cpp
@@ -14,7 +14,7 @@
 /// ----------------------------------------------------------------------------
 ///
 /// Persistence of Vision Ray Tracer ('POV-Ray') version 3.7.
-/// Copyright 1991-2016 Persistence of Vision Raytracer Pty. Ltd.
+/// Copyright 1991-2017 Persistence of Vision Raytracer Pty. Ltd.
 ///
 /// POV-Ray is free software: you can redistribute it and/or modify
 /// it under the terms of the GNU Affero General Public License as
@@ -6133,7 +6133,8 @@ DBL DensityFilePattern::EvaluateRaw(const Vector3d& EPoint, const Intersection *
                     Ex = max(0.0,Ex-(1.0/(DBL)Data->Sx/2.0)); // Adjustment to voxel center
                     Ey = max(0.0,Ey-(1.0/(DBL)Data->Sy/2.0));
                     Ez = max(0.0,Ez-(1.0/(DBL)Data->Sz/2.0));
-
+                    // FALLTHROUGH
+                case kDensityFileInterpolation_Trilinear_Shftd: // (11 vs 1) As means to provide backward shift left-forward-down compatibility
                     xx = Ex * (DBL)(Data->Sx);
                     yy = Ey * (DBL)(Data->Sy);
                     zz = Ez * (DBL)(Data->Sz);
@@ -6198,7 +6199,8 @@ DBL DensityFilePattern::EvaluateRaw(const Vector3d& EPoint, const Intersection *
                     Ex = max(0.0,Ex-(1.0/(DBL)Data->Sx/2.0)); // Adjustment to voxel center
                     Ey = max(0.0,Ey-(1.0/(DBL)Data->Sy/2.0));
                     Ez = max(0.0,Ez-(1.0/(DBL)Data->Sz/2.0));
-
+                    // FALLTHROUGH
+                case kDensityFileInterpolation_Tricubic_Shftd: // (12 vs 2) As means to provide backward shift left-forward-down compatibility
                     xx = Ex * (DBL)(Data->Sx);
                     yy = Ey * (DBL)(Data->Sy);
                     zz = Ez * (DBL)(Data->Sz);
--- a/source/core/material/pattern.cpp
+++ b/source/core/material/pattern.cpp
@@ -44,9 +44,6 @@
 
 #include <limits>
 #include <algorithm>
-#ifdef AVX2_INTRINSICS
-#include <immintrin.h>
-#endif
 
 #include "base/fileinputoutput.h"
 
@@ -6051,21 +6048,6 @@ inline DBL fblob(DBL v, DBL s)
   return 1.0/exp(v*s);
 }
 
-#ifdef AVX2_INTRINSICS
-// Needed only for AVX2 option of interpolation modes 3,4 and 5.
-inline void calcAllDiffsSqrd(double *aryd, double this_d, __m256i v_range)
-{  //---- Do all our dist calculations for axis
-   __m256d v_thisd    = _mm256_set1_pd(this_d);
-   __m256d v_rangexd1 = _mm256_cvtepi32_pd(_mm256_extracti128_si256(v_range,0b0));
-   __m256d v_rangexd2 = _mm256_cvtepi32_pd(_mm256_extracti128_si256(v_range,0b1));
-   __m256d v_diffx1   = _mm256_sub_pd(v_rangexd1,v_thisd);
-   __m256d v_diffx2   = _mm256_sub_pd(v_rangexd2,v_thisd);
-   __m256d v_diffx12  = _mm256_mul_pd(v_diffx1,v_diffx1);
-   __m256d v_diffx22  = _mm256_mul_pd(v_diffx2,v_diffx2);
-                        _mm256_store_pd(aryd,v_diffx12);
-                        _mm256_store_pd(&aryd[4],v_diffx22);
-}
-#else
 inline void calcAllDiffsSqrd(double *aryd, double this_d, size_t *ary)
 {  //---- Do all our dist calculations for axis at this point
    size_t n;
@@ -6076,7 +6058,6 @@ inline void calcAllDiffsSqrd(double *aryd, double this_d, size_t *ary)
        aryd[n]=tmpVal*tmpVal;
    }
 }
-#endif
 
 inline float intp3(float t, float fa, float fb, float fc, float fd)
 {
@@ -6117,9 +6098,7 @@ DBL DensityFilePattern::EvaluateRaw(const Vector3d& EPoint, const Intersection *
     DBL density = 0.0;
     DensityFilePattern::DensityFileDataStruct *Data;
     size_t k0, k1, k2, k3, i, j, k, ii, jj, kk;
-#ifndef AVX2_INTRINSICS
-    size_t rx[8],ry[8],rz[8];     // Not used if doing the avx2 method.
-#endif
+    size_t rx[8],ry[8],rz[8];
     DBL    rxd[8],ryd[8],rzd[8];
     size_t startidx, stopidx;
     DBL    runningval,strengthBias;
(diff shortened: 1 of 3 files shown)


c-lipka:
Whoa, this looks like it is a big one. A few preliminary comments just from browsing the docs and the list of commits:
- The "gain of around 1%" mentioned in commit 6f1f4a5 needs to be examined; given that this is claimed to be just the elimination of redundant code, no gain should be seen at all.
- Changing the alignment from vertex-based to voxel-based is a breaking change, and a backward compatibility mode should probably be provided. Besides, it needs to be examined whether the proper fix for the inconsistency is indeed to update all modes' alignment to match that of mode 0, or whether the alignment of mode 0 should be changed instead.
- The new density file interpolation modes, as fancy as they might seem, certainly deserve a thorough testing before incorporating them into official POV-Ray (for example, one thing I'd like to see examined carefully is whether the new modes mess with the average density of the file); I would therefore suggest to first publish them as a dedicated unofficial patch, or as part of an existing unofficial patch collection such as UberPOV or hgpovray.
- The gradient "pseudo-df3" modes will _never_ make it into official POV-Ray, since **they are not density file-based**. They should be implemented as separate patterns, and if I'm not mistaken they essentially already exist (gradient, spherical and cylindrical come to my mind); ease of use in conjunction with density file patterns can just as well be achieved by macros if necessary, at no rendering speed penalty.
- I find the docs overwhelming, with too much input for a casual first contact with the new modes.

wfpokorny:
(Is there a way to respond to comments in line on github...)
Yes, it kind of grew in size as I worked - then shrank some especially in pulling an avx2_intrinsics compile option after the gnu compiler caught up. The compiler mostly bettered my effort.      

The gain of 1% comes from range checking which was getting done twice. 'Redundant' was likely a poor choice of words as the compiler would eliminate truly redundant code. I did quite a few before and after performance tests with linux's time command. Did I miss your point?        

> Changing the alignment from vertex-based to voxel-based
Understand. Though I expect any compatibility mode will require a switch at some cost in performance. As to whether vertex or voxel centered is correct, I think the answer comes from the original interpolation 0 where we'd get only half the lower-left-front voxels inside the unit cube and we'd have a gap in the back-upper-right, if we moved 0 to be vertex centered.  

Testing. Yes agree & I'll be doing more myself. I leave my branch active indefinitely. I believe all the interpolation modes other than 0/no interpolation mess with the average density of the unit cube. I am surprised by any expectation otherwise here?   
  
> The gradient "pseudo-df3" modes will never make it into official POV-Ray
Hmm, I'm going to ask you to think more about this one... ;-) These were originally not planned. I was inspired by your own ideas some months back with respect to functions qualified by ranges. I am aware everything done with these can be done otherwise - I've done it that way. The complexity goes way up and performance down over these inbuilt flavors. Perhaps though these new blobbing interpolations and inbuilt helper interpolations should be done as some other isosurface aimed density_file like pattern leaving density_file as is - I'll think about that. New built in functions might be a path to better performance too. 

I believe the main reason isosurfaces don't get used by more casual users is coding the functions is hard. I've done it a lot and I still stumble all the time. My thinking was to wrap common functional operations required inside the density_file pattern so base shape definition and many of the operations which would otherwise require lots of function work could be done with the usual pattern modifiers most users know. 

> I find the docs overwhelming, with too much input for a casual first contact with the new modes.
:-) Sorry, I'm not the best writer and I throw in occasional semi-dyslexic nonsense too. I'll see if I can tighten it up some.

c-lipka:
> (Is there a way to respond to comments in line on github...)

(I'd be happy to even find a "quote this post" button..)

> The gain of 1% comes from range checking which was getting done twice. 'Redundant' was likely a poor choice of words as the compiler would eliminate truly redundant code. I did quite a few before and after performance tests with linux's time command. Did I miss your point?

Sorry, complete misunderstanding on my side there. I thought you meant a gain in density. It now dawns on me that you meant gain in speed, which is of course a good thing. :-)

>> Changing the alignment from vertex-based to voxel-based
>
> Understand. Though I expect any compatibility mode will require a switch at some cost in performance.

Not necessarily. One possibility would be to implement the backward-compatibility shift at the parser level, as an automatic addition of a translation to the pattern.

> I believe all the interpolation modes other than 0/no interpolation mess with the average density of the unit cube. I am surprised by any expectation otherwise here?

As a matter of fact, at least linear interpolation does indeed keep the average density stable, no matter the original pattern. I'm not sure about bicubic interpolation. But as a minimum requirement, I would expect even exotic interpolation modes to keep average density stable in the following cases (and the current interpolation modes do):
- All voxels have the same arbitrary value. Interpolation should leave no trace whatsoever in this case.
- An arbitrary plane divides space into two regions, each having a uniform arbitrary value. Interpolation should blur the boundary, but not boost nor reduce the average density.
- All voxels have one of two arbitrary values, arranged in an alternating fashion. Interpolation may even this out to a uniform value, but the average should again remain unaffected.

I suspect that depending on the interpolation formula, achieving the above may be as simple as taking the n-th root of the result (which obviously you can't do with the wave parameters-based approach).

>>  The gradient "pseudo-df3" modes will never make it into official POV-Ray
>
> Hmm, I'm going to ask you to think more about this one... ;-) These were originally not planned. I was inspired by your own ideas some months back with respect to functions qualified by ranges.

"Functions qualified by ranges" doesn't ring a bell in me.

> I believe the main reason isosurfaces don't get used by more casual users is coding the functions is hard. I've done it a lot and I still stumble all the time. My thinking was to wrap common functional operations required inside the density_file pattern so base shape definition and many of the operations which would otherwise require lots of function work could be done with the usual pattern modifiers most users know.

The job of POV-Ray's rendering engine, and even the SDL syntax, is _not_ to cater to a specific "style" of modelling - be it mesh-based, isosurface-based, blob-based, or even CSG-based. Its job is to provide flexible tools that are useful for as many styles as possible. Provision of style-specific tools is only warranted if the existing tools can't fill that niche, or if a specific tool will provide a mind-boggling boost in performance.

Adapting POV-Ray's existing tools to be _easy_ to use with a specific style is what we have macros for.

wfpokorny:
>>>     Changing the alignment from vertex-based to voxel-based

>>  Understand. Though I expect any compatibility mode will require a switch at some cost in performance.

> Not necessarily. One possibility would be to implement the backward-compatibility shift at the parser level, as an automatic addition of a translation to the pattern

OK. I'll let this thought rattle around in my head for a while.

>> I believe all the interpolation modes other than 0/no interpolation mess with the average density of the unit cube. I am surprised by any expectation otherwise here?

> As a matter of fact, at least linear interpolation does indeed keep the average density stable, no matter the original pattern. I'm not sure about bicubic interpolation. But as a minimum requirement, I would expect even exotic interpolation modes to keep average density stable in the following cases (and the current interpolation modes do):

> * All voxels have the same arbitrary value. Interpolation should leave no trace whatsoever in this case.

> * An arbitrary plane divides space into two regions, each having a uniform arbitrary value. Interpolation should blur the boundary, but not boost nor reduce the average density.

> * All voxels have one of two arbitrary values, arranged in an alternating fashion. Interpolation may even this out to a uniform value, but the average should again remain unaffected.

> I suspect that depending on the interpolation formula, achieving the above may be as simple as taking the n-th root of the result (which obviously you can't do with the wave parameters-based approach)

OK.If I find myself in the mood, I'll try some test cases to see where we really are.  Practically folks should only count on 0 representing their original data well - unless they buffer with 0.0 values on the sides then perhaps 1 is OK.  Maybe 2 actually works out for the unit cube density, but with the ringing & clipping(>1 values wrap back to 0) any local density in the unit cube is suspect.

Now wondering if trying to maintain the unit cube density is why interpolations 1 & 2 were left wrapping the unit cube...  Might be better maintaining any measured unit cube's density that way, but at the cost of badly representing local density on the edges.

> "Functions qualified by ranges" doesn't ring a bell in me.

I was thinking the blob alternative you and Jérôme proposed some time ago.

> The job of POV-Ray's rendering engine, and even the SDL syntax, is not to cater to a specific "style" of modelling - be it mesh-based, isosurface-based, blob-based, or even CSG-based.  Its job is to provide flexible tools that are useful for as many styles as possible. Provision of style-specific tools is only warranted if the existing tools can't fill that niche, or if a specific tool will provide a mind-boggling boost in performance.

> Adapting POV-Ray's existing tools to be easy to use with a specific style is what we have macros for

Alright. I'll plan on removing the helper interpolations 6-12. On waking this morning I was thinking the performance boost comes mostly with 10-12. Those 'helper functions' can more flexibly be implemented as additional built in functions. 6-9 can be done other ways at decent performance. In dropping 6 we'll lose too the back door to get the df3 dimensions I added as part of interpolation 6.   

On performance in general... Please keep in mind much less than "mind-boggling" changes in performance greatly impact what can practically be done with isosurfaces.

c-lipka:
> In dropping 6 we'll lose too the back door to get the df3 dimensions I added as part of interpolation 6.

I agree that a function giving the resolution of a df3 might be useful, but that shouldn't be by means of a  "back door" using a quirky side effect of a quirky hack. Besides, a "front door" solution to that problem has already been implemented for image-based pigments, overloading the "max_extent()" function, so for consistency that same mechanism should be used for df3-based stuff.

wfpokorny:
>> In dropping 6 we'll lose too the back door to get the df3 dimensions I added as part of interpolation 6.

> I agree that a function giving the resolution of a df3 might be useful, but that shouldn't be by means of a "back door" using a quirky side effect of a quirky hack. Besides, a "front door" solution to that problem has already been implemented for image-based pigments, overloading the "max_extent()" function, so for consistency that same mechanism should be used for df3-based stuff.

I missed the max_extent image resolution addition in 3.7. On looking at the documentation, do you agree the example provided in:

//---
#declare MatteImage =
pigment {
  image_map {
    jpeg "YourImage.jpg"
    once
    }
  }

#declare Resolution = max_extent ( MatteImage );

box { 0, Resolution
  pigment { MatteImage  }
  // avoid the scale by warning (z direction)
  #declare Resolution = Resolution + <0, 0, 1>;
  // some scale factor to fit your needs
  scale Resolution*ScaleFactor
}
//---

would be better as:

//---
#declare MatteImage = pigment { image_map { "YourImage.jpg" } }

#declare Resolution = max_extent ( MatteImage );

#declare MatteNormalizationScale =
<min(1,Resolutione.x/Resolutione.y),min(1,Resolutione.y/Resolutione.x),1>;

box { 0, <1,1,0.001>
  pigment { MatteImage  }
  scale MatteNormalizationScale // Fit matte to unit square
  scale MatteSceneScale         // Size matte for scene
}
//---

wfpokorny:
I have a version where I've pulled out the assist interpolations 6-12, which I'll push here after more testing. 

In scenes where I was using the standard pattern modifiers on both the base df3-isosurface and an assist interpolation in combination for the isosurfaces are running about 2x slower.

The attached images show a test which was 133% slower using existing function based mechanisms (images are a little different due - complications).

dualturb

The worst case degrade happens due having to set up three functions for x, y and z based upon gradient patterns as the way to use the standard pattern modifiers. There are also bounding issues with the current methods which make doing this sort of dual turbulence trick harder.

Better, function integration with the standard pattern modifiers, while allowing them to be applied to the working space for a passed function, would help and be applicable to more that the density_file pattern. Only a vague notion in my mind at the moment.

c-lipka:

wfpokorny:
I've additional ideas here. Closing the pull request while keeping the branch for my work space.

PR #80, opened 2016-08-22
Adding df3, density_file query range support to max_extent.

wfpokorny:
Enables users to determine the x,y,z ranges, returned as a vector, for
density_file based pigments matching a similar function for image maps.

### DIFF
--- a/source/parser/parser_expressions.cpp
+++ b/source/parser/parser_expressions.cpp
@@ -1212,6 +1212,7 @@ void Parser::Parse_Num_Factor (EXPRESS& Express,int *Terms)
                 case MAX_EXTENT_TOKEN:
                     GET (LEFT_PAREN_TOKEN);
                     EXPECT
+
                         CASE (OBJECT_ID_TOKEN)
                             Object = reinterpret_cast<ObjectPtr>(Token.Data);
                             if ( Object )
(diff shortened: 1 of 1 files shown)


source/parser/parser_expressions.cpp
```diff
@@ -1222,16 +1222,22 @@ void Parser::Parse_Num_Factor (EXPRESS& Express,int *Terms)
                         // JN2007: Image map dimensions:
                         CASE (PIGMENT_ID_TOKEN)
```
c-lipka: Currently this only works for `pigment`, but I think we also want it to work with `density`, so I suggest to replace the above with `CASE2 (PIGMENT_ID_TOKEN, DENSITY_ID_TOKEN)`.

source/parser/parser_expressions.cpp
```diff
@@ -1222,16 +1222,22 @@ void Parser::Parse_Num_Factor (EXPRESS& Express,int *Terms)
...
                             Pigment = reinterpret_cast<PIGMENT *>(Token.Data);
-                            if(Pigment->Type != BITMAP_PATTERN)
+                            if((Pigment->Type != BITMAP_PATTERN) && (Pigment->Type != DENSITY_FILE_PATTERN))
```
c-lipka: While you're at it, can you do me a favour and rearrange the structure from `if (!A && !B) { ... } else if (A) { ... } else if (B) { ... }` to `if (A) { ... } else if (B) { ... } else { ... }`?

PR #81, opened 2016-08-27
Allowing the use of pigment_map in density blocks.

wfpokorny:
In particular this enables the df3, image plane stacking done for clouds
and such to be done on the fly within povray. In general makes density
more flexible and it does it while doing all the color handling with no
additional overhead.


c-lipka:
What you've tampered with there is a portion of the parser code that I consider too fragile to tamper with.

I suspect that the thing you're trying to achieve might be achieved more easily and far less intrusive by allowing the use of image_map inside a density statement.

wfpokorny:
Thanks for your take. Closing. 

I'll submit a new pull from another branch - IF - I can quickly figure out how to do just the image_map...

PR #82, opened 2016-08-28
Adding image_map support to density.

wfpokorny:
A less risky alternative to adding pigment_map support to
density(#81-closed). This method also enables image slices to define a
density in a way other than the external image file slices to df3
methods. The densities would be defined with image files and those
would be used in a density_map.

### DIFF
--- a/source/parser/parser_materials.cpp
+++ b/source/parser/parser_materials.cpp
@@ -8,7 +8,7 @@
 /// @parblock
 ///
 /// Persistence of Vision Ray Tracer ('POV-Ray') version 3.7.
-/// Copyright 1991-2016 Persistence of Vision Raytracer Pty. Ltd.
+/// Copyright 1991-2017 Persistence of Vision Raytracer Pty. Ltd.
 ///
 /// POV-Ray is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


wfpokorny:
DO NOT pick this up yet... I think I see a parser issue. Digging.

c-lipka:
Density maps indeed cannot be assigned to a variable, but neither can pigment maps; `density{}` blocks, on the other hand, should be assignable just as fine as pigments.

wfpokorny:
I am perhaps not understanding. While the attached test code is written against a POV-Ray version where I am allowing both image_map and pigment_map in density {}, assigning a pigment_map to a variable and using that variable in downstream pigment definitions is something I do routinely on setting up a pigment_map. 

I tried too to use a density {} within pigment {} a few different ways, but I could not get the code past the parser. I agree density {} would likely otherwise work as a pigment. 
   
[tmp.pov.txt](https://github.com/POV-Ray/povray/files/445242/tmp.pov.txt)

c-lipka:

wfpokorny:
Closing. Always a personally unused subset of a pigment_map+image_map in density branch I do use. Work of keeping this current long ago exceeded its value as a trivial edit...

PR #84, opened 2016-09-02
Porting Jerome Grimbert's lemon object to master.

wfpokorny:
Initial effort at a port of the lemon object from Hgpovray. My testing looks
ok, but please review.


c-lipka:
My colleague Mr. AppVeyor says I shouldn't bother to review the changes just yet, since you might want to do a bit more work on it ;)

My diagnosis is that the new files need to be added to the Visual Studio projects. (Unlike the Unix build process, the Unix build process can't auto-add new files.) It shouldn't be too difficult: Just open the files `windows/vs10/povcore.vcxproj` and `windows/vs10/povcore.vcxproj.filters` in a text or XML editor, look for one of the sibling files (e.g. `sphere.cpp` and `sphere.h`), and add corresponding entries for the new files.

c-lipka:
In other news, can you please update `POV_RAY_PRERELEASE` in `source/base/version.h` by a notch, and also adjust `unix/VERSION` accordingly? That would be great.

You can automate this task by copying `tools/git/hooks/pre-commit` to `.git/hooks` in your local repo, causing that script to be run whenever you commit any changes. (UberPOV, by the way, has a similar script.)

wfpokorny:
Thanks for the pointers Christoph. I'll look at the issues, but not until next week as I'm busy with a family event this weekend.

PR #86, opened 2016-09-09
Feature/new black hole types

wfpokorny:
Only one new type of 1 at this time. See code and commit comments. 

[NewBlack_Hole_Type1.pov.txt](https://github.com/POV-Ray/povray/files/464862/NewBlack_Hole_Type1.pov.txt)
newblack_hole_type1

Files attached. 

Christoph, Just realized I forgot to turn on the execute bits of the commit hook script. I'll update this pull request after the checks complete.

### DIFF
--- a/source/core/material/warp.cpp
+++ b/source/core/material/warp.cpp
@@ -63,7 +63,6 @@ enum BlackHoleType
 {
     kBlackHoleSpherical_Original=0,
     kBlackHoleSpherical_Clamped,
-    kBlackHoleSpherical_WhatMightBeUseful
 };
 
 /*****************************************************************************
@@ -208,10 +207,10 @@ bool BlackHoleWarp::WarpPoint(Vector3d& TPoint) const
             // the surface of the spherical region for the inward pulling black hole.
             //
             // When inverted (pushing outward), clamping S to 1.0 is enough because
-            // well never look further inward than the origin of the spherical region
+            // we'll never look further inward than the origin of the spherical region
             // for the point to push outward. Good.
             //
-            // When pulling inward, clamping S to 1.0 OK for effects, but it does
+            // When pulling inward, clamping S to 1.0 is OK for effects, but it does
             // not prevent the code from reaching outside the spherical region for
             // the point to pull. The type below does and provides for a more realistic
             // and controllable inward pulling black_hole.
--- a/source/core/material/warp.cpp
+++ b/source/core/material/warp.cpp
@@ -59,6 +59,13 @@ namespace pov
 
 const DBL COORDINATE_LIMIT = 1.0e17;
 
+enum BlackHoleType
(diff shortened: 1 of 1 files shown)


wfpokorny:
Hmm. Apparently bumping the version number is causing a checking fail. Are we perhaps colliding on the differing version numbers themselves?

wfpokorny:
Adding reference and another pov example. 

Newsgroup discussion at: 
http://news.povray.org/povray.beta-test/thread/%3C57d333e2%241%40news.povray.org%3E/

[ExampleAsPigment_0_3_9.pov.txt](https://github.com/POV-Ray/povray/files/465912/ExampleAsPigment_0_3_9.pov.txt)

c-lipka:

c-lipka:
It is a mystery to me why one of your pull requests picked up the very newest updates to `appveyor.yml` but others like this one don't. Can you try whether re-basing this pull request onto the newest master fixes the issue?

source/core/material/warp.cpp
```diff
@@ -59,6 +59,13 @@ namespace pov
 
 const DBL COORDINATE_LIMIT = 1.0e17;
 
+enum BlackHoleType
+{
+    kBlackHoleSpherical_Original=0,
+    kBlackHoleSpherical_Clamped,
+    kBlackHoleSpherical_WhatMightBeUseful
```
c-lipka: Can we remove this?

wfpokorny: Yes, I can remove those. I have an appointment this morning so the update will come later today.

c-lipka: For clarity, I specifically meant the `kBlackHoleSpherical_WhatMightBeUseful` value. It's not used, as far as I can see, so it shouldn't be there. The others correspond to actual implementations, so I guess they're ok.

wfpokorny: Thanks for the clarification. I was going to yank all of them. I'd originally planned to do at least one more type, but never got around to it. Commit for the change just pushed.

This req is today against the master branch. A time prior I tried to move another pull request over to a different branch via github and the process failed...  I am going to leave the req aimed as is hoping you know how to make this branch change work. Let me know if the branch change is trouble for you too and I'll create a new pull request against 3.8.0 for this update.

c-lipka:
Haven't figured out the secret of branch changes either. Not in GitHub's pull requests at any rate. I'd have tried a rebase of the single commit (when it was still single) outside GitHub.

I suggest you hold off on the branch change until a decision is made whether we actually want to add this to 3.8.0.

wfpokorny:
Given this is now aimed at 3.8.1 I'm going to close this pull request for the moment. I have other black_hole types I'd like to try and I don't want to be triggering automated build and checks as I play.

PR #88, opened 2016-09-12
Implemented `potential` pattern and isosurface `polarity` keyword.

c-lipka:


wfpokorny:
OK. I woke up this morning with a thought on why the threshold option isn't working in my testing.

In pattern.cpp we need to copy subtractThreshold too when copying the object pointer. 
    if (obj.pObject)
    {
        pObject = Copy_Object(obj.pObject);
        subtractThreshold = obj.subtractThreshold;  // <--- Add this 
    }

Fixes threshold for what I'm testing in any case. 

Probably something similar happen with polarity. Perhaps too the isosurface container shape?

PR #91, opened 2016-09-14
Feature/new parser warn on color float

wfpokorny:
Further discussion in the newsgroup thread at: 

http://news.povray.org/povray.binaries.images/thread/%3C57d8340d%241%40news.povray.org%3E/

My scene test file: 

[testParser.pov.txt](https://github.com/POV-Ray/povray/files/471733/testParser.pov.txt)

### DIFF
--- a/source/parser/parser_expressions.cpp
+++ b/source/parser/parser_expressions.cpp
@@ -2463,7 +2463,7 @@ void Parser::Parse_Colour (RGBFTColour& colour, bool expectFT)
 {
     EXPRESS Express;
     int Terms;
-    bool old_allow_id = Allow_Identifier_In_Call;
+    bool old_allow_id = Allow_Identifier_In_Call, sawFloatOrFloatFnct;
     Allow_Identifier_In_Call = false;
 
     /* Initialize expression. [DB 12/94] */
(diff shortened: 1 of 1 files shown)


source/parser/parser_expressions.cpp
```diff
@@ -2693,16 +2693,28 @@ void Parser::Parse_Colour (RGBFTColour& colour, bool expectFT)
...
+                if ((Token.Token_Id==FLOAT_FUNCT_TOKEN) || (Token.Token_Id==FUNCT_ID_TOKEN))
+                    sawFloatOrFloatFnct = true;
+                else
+                    sawFloatOrFloatFnct = false;
                 if (expectFT)
                     Terms = 5;
                 else
                     Terms = 3;
                 Parse_Express(Express,&Terms);
+                // Note: Doubt expectFT test can fail. Was intent for float to error out, but we've not been?
```
c-lipka: Please reword this comment; I have difficulty deciphering it ;)

c-lipka:
Hm... reviewing this change seems to be just about as complex a task as to actually develop it.

From my first look at it, I found about half a dozen reasons why it couldn't possibly work, only to find upon double-checking that each of them was based on a misconception on my part, and the code may work properly after all. I guess I'll want to take a second look at it later, when I have digested the first helping of examining it.

I hate the parser! ;)

wfpokorny:
Thanks for reviewing. I'll clean up - or delete - that comment when I'm back at my povray computer tomorrow morning. If I correctly deciphered what is going on, the conditional below the comment will never be true.   

Yep! I found this update a whole lot harder to do than I imagined it would be. The parser is a tangle. 

I did attach my test scene if you want to start there testing things.  

On a completely different subject. I tried and failed early today to figure out the reason the threshold isn't working with the new potential pattern. I can see the parser setting the value. I can hard code the call the blob->getpotential (? I'm not remembering the names exactly) and see the adjustment run. Seems somewhere in between the parser and the call something goes wrong. The code there is really C++ and I am still climbing a big learning curve. I need to get a good C++ debugger going too for questions like this. Anyway, I have doubts I'm going to figure it out soon myself. 

Have a good night.

PR #93, opened 2016-09-18
I/O fixes + misc.

BentSm:
A few small fixes and adjustments -- some fixes for a couple of file I/O issues, addition of explicit casts to bool in a few places to make the code C++11-compatible, and an addition to .gitignore (unix/config/compile, created by automake).

(I know that this covers multiple issues, but the changes are so minor that I wasn't sure that it was worth the extra pull requests; if you'd rather have me split this up, I'll be glad to do so, though.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -186,6 +186,7 @@ $RECYCLE.BIN/
 /source/Makefile.in
 /unix/Makefile.am
 /unix/Makefile.in
+/unix/config/compile
 /unix/config/config.guess
 /unix/config/config.sub
 /unix/config/depcomp
--- a/source/base/fileinputoutput.cpp
+++ b/source/base/fileinputoutput.cpp
@@ -244,6 +244,8 @@ OStream::OStream(const UCS2String& name, unsigned int Flags) : IOBase(name), f(N
 
         if (f != NULL)
         {
+            fail = false;
+
             if((Flags & append) != 0)
             {
                 if(!seekg(0, seek_end))
--- a/source/base/fileinputoutput.cpp
+++ b/source/base/fileinputoutput.cpp
@@ -309,6 +309,8 @@ void OStream::printf(const char *format, ...)
 
 IStream *NewIStream(const Path& p, unsigned int stype)
 {
+    IFileStream *stream;
+
     if (!PlatformBase::GetInstance().AllowLocalFileAccess(p(), stype, false))
     {
         string str ("IO Restrictions prohibit read access to '") ;
@@ -317,12 +319,19 @@ IStream *NewIStream(const Path& p, unsigned int stype)
         throw POV_EXCEPTION(kCannotOpenFileErr, str);
     }
 
-    return new IFileStream(p().c_str());
+    stream = new IFileStream(p().c_str());
+    if (!(*stream))
+    {
+        delete stream;
+        return NULL;
+    }
+    return stream;
 }
 
 OStream *NewOStream(const Path& p, unsigned int stype, bool sappend)
 {
     unsigned int Flags = IOBase::none;
+    OStream *stream;
 
     if(sappend)
         Flags |= IOBase::append;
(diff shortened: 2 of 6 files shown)


source/base/fileinputoutput.cpp
```diff
@@ -317,12 +319,19 @@ IStream *NewIStream(const Path& p, unsigned int stype)
...
     }
 
-    return new IFileStream(p().c_str());
+    stream = new IFileStream(p().c_str());
+    if (!(*stream))
+    {
+        delete stream;
+        return NULL;
+    }
+    return stream;
```
c-lipka: It could be argued that the error is not in these functions, but in the callers' failure to test for errors. I think the proper first step to address this issue is to decide on, and formally specify, the intended behaviour of the functions; before that has happened, any code changes are moot.

Given that the functions already report some error situations by throwing an exception, it can be argued that the same mechanism should be used for other error situations as well.

.gitignore
```diff
@@ -186,6 +186,7 @@ $RECYCLE.BIN/
 /source/Makefile.in
 /unix/Makefile.am
 /unix/Makefile.in
+/unix/config/compile
```
c-lipka: The cleanup of automake artifacts in `unix/prebuild.sh` should probably be updated accordingly (search for `config.guess`)

source/base/image/hdr.cpp
```diff
@@ -132,7 +132,7 @@ void ReadOldLine(unsigned char *scanline, int width, IStream *file)
 
         // NB EOF won't be set at this point even if the last read obtained the
         // final byte in the file (we need to read another byte for that to happen).
-        if(*file == false)
+        if((bool)(*file) == false)
```
BentSm: Will do.

.gitignore
BentSm: Will do.

source/base/image/hdr.cpp
BentSm: This commit definitely has issues.  My apologies.

The only thing left that appears to be any issue is this specific line, and not because of C++11.  (I'm really not sure what I was thinking...)

Anyways, I am going to replace it with a commit to handle just that one issue, since the rest doesn't make very much sense.

source/base/fileinputoutput.cpp
BentSm: I had chosen this method after finding that much (if not all) of the code downstream from these functions uses comparison to NULL to detect failure (and throw an exception); e.g.:

* source/base/image/image.cpp:3943 (via ????)
* source/backend/scene/backendscenedata.cpp:292 (via NewOStream)
* source/backend/scene/backendscenedata.cpp:303 (via NewOStream)
* source/frontend/imageprocessing.cpp:181 (via NewOStream)
* source/frontend/processoptions.cpp:81 (via OpenFileForRead)
* source/frontend/processoptions.cpp:373 (via OpenFileForWrite)
* source/parser/parser.cpp:6499 (via Locate_File)
* source/parser/parser.cpp:10937 (via ReadFile)
* source/parser/parser_materials.cpp:1443 (via Locate_File)
* source/parser/parser_tokenizer.cpp:2608 (via Locate_File)
...

I'm not entirely sure, but it seems that such code may have been relying on the cast from `IOBase` to `const void *` (in analogy to the standard C++ I/O); unfortunately, it uses `IOBase *`, which doesn't work the same.  I chose to modify NewIStream / NewOStream since all (as far as I can tell) of such 'confused'(?) code passes through those functions, and this would restore the expected function.

I'd be willing to try to convert things to use exceptions directly (or whatever), but it may take some time.

Let me know.

.gitignore
BentSm: Fixed.

source/base/image/hdr.cpp
BentSm: Done.

source/base/fileinputoutput.cpp
BentSm: OK, it looks like my theory about the cast was wrong...
Prior to commit 051b674, returning NULL on failure actually was the behavior of NewIStream / NewOStream (and, as far as I can tell, had been since version 3.5).

BentSm:
Closing and splitting.

PR #102, opened 2016-10-01
Fix 4D vector parsing anomaly

BentSm:
When expecting a 4D vector, the parser would erroneously treat a 3D vector as if it had 4 components, despite only setting 3 (so that the fourth component would retain whatever value was in that memory location).  (More precisely, Parse_Num_Factor would fail to update `*Terms` when parsing a 3D vector token if `*Terms>3`.)  This would cause, e.g., `julia_fractal{2/9*x}` to get parsed as if one had written `julia_fractal{2/9*<1,0,0,9>}`.

This fixes that issue.  (I also removed a nearby comment that no longer makes much sense.)

### DIFF
--- a/source/parser/parser_expressions.cpp
+++ b/source/parser/parser_expressions.cpp
@@ -1256,8 +1256,7 @@ void Parser::Parse_Num_Factor (EXPRESS& Express,int *Terms)
                I haven't yet figured out what to do if it was expecting
                a COLOUR value with Terms>3
             */
-            if(*Terms < 3)
-                *Terms = 3;
+            *Terms = 3;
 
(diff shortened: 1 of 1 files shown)


c-lipka:
Thanks for this fix!

Just for the records: Are you aware of any place where this had been reported previously (GitHub issue, the old FlySpray tracker, one of the newsgroups, or anything along those lines)?

BentSm:
No, but it's possible; I didn't check.

c-lipka:
If you happen to have the time, that would be great.

BentSm:
After a little looking, it does not appear to have been reported.

PR #104, opened 2016-10-01
Changing two bug tracking refs to github issues.

wfpokorny:
Changed two unix file references from http://bugs.povray.org/
to https://github.com/POV-Ray/povray/issues

### DIFF
--- a/unix/install.txt
+++ b/unix/install.txt
@@ -3,8 +3,9 @@ POV-Ray 3.7 for UNIX/Linux - INSTALL file
 
 Compiling and installing POV-Ray for Unix from sources.
 
-Please report bugs via our bugtracker at http://bugs.povray.org/ or to our forum
-at news://news.povray.org/povray.unix/ or http://news.povray.org/povray.unix/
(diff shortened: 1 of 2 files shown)


c-lipka:
Careful when editing `prebuild.sh` -- portions of it end up in makefiles, where tab characters do matter, so you _must not_ convert tabs to spaces in that file.

That said, I already have some updates to the Unix build process in the pipeline, so it may just be a waste of your time to fix this pull request.

wfpokorny:
Thanks. I didn't think about tabs. While I ran prebuild.sh after the change, I also did not try a configure and make...   

Please leave this open a little while. I am going to play with updating this to build in my own space after which I want to see if the versions I get to build locally will also pass the checks on github too. 

At the moment looks like my normal editor tab-in and tab save(8 spaces) for makefiles is making some other changes during the edit, so I'll close this pull req no matter. 

Please plan on this pull request being soon closed by me and the github issues updates to unix/install.txt and unix/prebuild.sh being yours to do with your other build updates.

wfpokorny:
Closing per earlier comment.

PR #106, opened 2016-10-02
Minor fix to hdr.cpp

BentSm:
An invalid comparison of IStream to bool in hdr.cpp prevented compilation.  This fixes that.


c-lipka:
For the records, please let us know what compiler choked on the previous version of the code.

BentSm:
Apple LLVM version 8.0.0 (clang-800.0.38) with -std=c++11.

PR #109, opened 2016-10-07
Improving unix console messages with SDL display use.

wfpokorny:
The end render message:  Press a key or click the display to continue...
for unix versions has long been inaccurate because you cannot click any
key, but rather only a few specific ones.  Pauses with 'p' also gave no
clue in the paused message to use 'p' to continue.  Update somewhat
complicated for being only text message changes due no-display runs or
compiles without SDL installed - and so no-display - not picking up
general key press events.

### DIFF
--- a/unix/disp.h
+++ b/unix/disp.h
@@ -11,7 +11,7 @@
 /// @parblock
 ///
 /// Persistence of Vision Ray Tracer ('POV-Ray') version 3.7.
-/// Copyright 1991-2014 Persistence of Vision Raytracer Pty. Ltd.
+/// Copyright 1991-2016 Persistence of Vision Raytracer Pty. Ltd.
 ///
 /// POV-Ray is free software: you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


wfpokorny:
Thanks for the review. I've made the changes and pushed them as a rebase to master with the updates flattened into the previous commit. 

- Fixed the one conditional test longer than 120 characters to which I added a test, but I didn't attempt to fix all the other lines over 120 chars already in the disp_sdl.cpp file. 

- The reason to allow 'p' to continue after the pause caused by '+P' at the end of a render was that the  
"==== [Paused... Press p to resume] ====..." screen message was used for both the in render pause and the '+P' one, but 'p' wasn't previously used as a resume key press for the latter pause state. Allowing 'p' to resume at the end render pause I thought both consistent and, code wise, the simplest way to make the screen messages and behavior consistent.   

- Added the spaces requested in unixconsole.cpp.

PR #116, opened 2016-10-13
Feature unix terminal

eblanca:
With this patch my povray executable can detect the terminal width and adapt the wrapping limit.

Please, review this code because my c++ knowledge is poor.
In particular, in order to react to a WINdowCHange event, I had to move the session object, making it global, and this is a ugly solution indeed.

Further, into the 'InitInfo' method in renderfronted.cpp, it should be very nice to change NUMBER_OF_AUTHORS_ACROSS and make its value a function of actual terminal width.
I would like to do it by myself but in this module I have no access to the 'session' object (for GetConsoleWidth() ).

### DIFF
--- a/vfe/unix/unixconsole.cpp
+++ b/vfe/unix/unixconsole.cpp
@@ -43,11 +43,554 @@
 #include <termios.h>
 #include <unistd.h>
 #include <sys/select.h>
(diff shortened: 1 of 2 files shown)


eblanca:
No, in some way I did a mess with this fork. Sorry.

PR #117, opened 2016-10-13
Feature unix terminal

eblanca:
With this patch my povray executable can detect the terminal width and adapt the wrapping limit.

Please, review this code because my c++ knowledge is poor.
In particular, in order to react to a WINdowCHange event, I had to move the session object, making it global, and this is a ugly solution indeed.

Further, into the 'InitInfo' method in renderfronted.cpp, it should be very nice to change NUMBER_OF_AUTHORS_ACROSS and make its value a function of actual terminal width.
I would like to do it by myself but in this module I have no access to the 'session' object (for GetConsoleWidth() ).


eblanca:
Thank you for your endless suggestions, I have updated my fork, and now it should be compliant with your guidelines.
About the header, with a fast review I see it is almost the same (and in the same path) in linux, freebsd (checked in a vm), openbsd (checked in a vm) and macos x (apple docs). In solaris 10/11 the 'ioctl' requires both unistd.h and stropts.h but I couldn't verify that.

c-lipka:
> Thank you for your endless suggestions, I have updated my fork, and now it should be compliant with your guidelines.

Thanks. Unfortunately, your code still doesn't run properly on Mac OS X. I'm no Mac OS X expert, so I have no idea what's wrong there.

> About the header, with a fast review I see it is almost the same (and in the same path) in linux, freebsd (checked in a vm), openbsd (checked in a vm) and macos x (apple docs). In solaris 10/11 the 'ioctl' requires both unistd.h and stropts.h but I couldn't verify that.

I would consider your Solaris research results evidence that `ioctl` may _not_ be all that standardized, and would therefore strongly suggest adding a test to `unix/configure.ac` to check for presence of the header file and necessary functions and macros, and add preprocessor safeguards to the relevant portions of the code, falling back to a constant terminal width if necessary.

eblanca:
Updated configure.ac and unixconsole.cpp

c-lipka:
The code still doesn't run when compiled on Mac OS. Could it be that the signal handler is hooked up and triggered before `gSession` is properly initialized?

eblanca:
At least I see with travis both under ubuntu and under mac os the build completes successfully.
There is indeed something subtle with threads.

c-lipka:
Try initializing `gSession` to `NULL` in `main()` before hooking up the signal handler.

eblanca:
Really, I don't get what can be wrong.
The error messages at the end of travis builds are:
```
Timed out waiting for worker thread startup

libc++abi.dylib:
  terminating with uncaught exception of type
  boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::lock_error> >:
    boost: mutex lock failed in pthread_mutex_lock: Invalid argument
```
and
```
Timed out waiting for worker thread startup

pure virtual method called

povray: /usr/include/boost/thread/pthread/mutex.hpp:47: boost::mutex::~mutex(): Assertion `!pthread_mutex_destroy(&m)' failed.

terminate called without an active exception
```
It seems they both are thread related, as boost is feeding pthread APIs with non valid arguments coming from my code.

I see into unixconsole.cpp there is this mutex called gSignalMutex used into two different functions and both lock it, when noone unlocks it. May this be a flaw?

c-lipka:
> I see into unixconsole.cpp there is this mutex called gSignalMutex used into two different functions and both lock it, when noone unlocks it. May this be a flaw?

Those mutexes are probably fine. They're using boost scoped locks, which self-unlock when going out of scope.

I'm as puzzled as you are. My hope was that the "pure virtual method called" might have been caused by an attempt to invoke a virtual method on an uninitialized pointer, but apparently that's not the answer to the riddle.

eblanca:
Trying to re-create the same enviroment as Travis, I've built an ubuntu virtual machine using their 12.04.5 iso, setting up thing just as the failing test but my povray executable does not crash.
Same libraries installed (libpng, boost, exr and such) and same enviroment variables (-fno-fast-math passed to g++) and the check goes successfully.

c-lipka:
The crash is probably related to the binary not running in a plain terminal window, but rather as a background task with the output redirected to a log (or at least that's how I presume Travis CI does it).

eblanca:
Mmmmmhhhhhh... right.
I managed to run "make check" in background (no terminal window) and indeed the ioctl call fails returning -1 and the little struct which should contain the terminal size is filled with random values. Sure, this case deserves to be handled in some way.
But still no crashes.

eblanca:
Although this latest commit makes Travis happy, I'm still not satisfied.
Placing the column setup into unixconsole.cpp the way it is now, this allows some strings to be formatted before that, resulting in wrong format.
I think I will fix this tonight.

eblanca:
It doesn't like this either.
Further, I see I'm moving towards a wrong target, using the vfeUnixSession class, instead wrapping is done inside TextStreamBuffer class, and I should instead modify that class (maybe).

eblanca:
Closing this PR, I'll open a new one.

Issue #122, opened 2016-10-15
Port of FS306 - finish subsurface block before global_settings subsurface block causes crash


wfpokorny:
http://bugs.povray.org/task/306

-------------------------------------
Details:

The following scene causes a crash:

```
sphere {
  <0,0,0>, 1
  finish { subsurface { translucency 1.0 } }
}

global_settings {
  subsurface { }
}
```

wfpokorny:
See pull req #123.

c-lipka:
For the records, the crash is caused by the subsurface computation code relying on the presence of a data block with some pre-computed data; but this data block is only created if subsurface is enabled via the "subsurface{}" global setting.

c-lipka:
Any solution to this issue is subject to the following constraint:

It is the intended behaviour of the finish subsurface block that you _can_ use it in a scene that does not enable subsurface scattering at all, in order to facilitate definition of materials that can be used regardless of whether subsurface scattering is enabled or not; no error should be produced, and as a matter of fact not even a warning.

The true root cause of the issue is that the pre-computed subsurface data block is intended to be created when the scene has been fully parsed and it is clear whether the scene will be using subsurface scattering or not. That is currently not the case, and instead pre-computation is invoked as soon as an object is "placed" in the scene.

I'll dig into this matter myself to figure out how to best fix this, but feel free to also continue investigating and making suggestions.

wfpokorny:
Not opposed to better solutions. 

I only saw the crash where the global block subsurface followed an object using subsurface. 

For users who do not specify the subsurface block in a global block, but only use subsurface on one or more objects - the feature simply doesn't work. It doesn't work despite the user getting a parser warning telling them they are using the experimental, subsurface feature. 

I suggested the immediate parser fix more for the latter reason than the crash.

c-lipka:
For the records: Since subsurface scattering is still tagged as an experimental feature, I'll call this something we don't want to burden ourselves with in preparation for v3.8.0.

PR #123, opened 2016-10-15
Fix for subsurface crash. Issue #122 and FS306.

wfpokorny:
Added a check in the parser that subsurface scene data true on
seeing finish subsurface block.


c-lipka:
This approach doesn't cut it; see issue #122 for details.

Since any working solution effectively needs to start from scratch anyway, I'm closing this pull request.

PR #130, opened 2016-10-23
Fix promotion bug

BentSm:
This fixes the bug reported in http://news.povray.org/3c4bb31d$1@news.povray.org.  (This also does some general cleanup of source/parser/parser_expressions.cpp.)

(Promotion was occurring before it was needed, causing (sub-)expressions to produce different results depending on context.  For instance, floats were being directly promoted by `Parse_Num_Factor`, causing, e.g.: `<1,0,0>+1*t` to parse as `<1,0,0>+<1,1,1>*t` and thence `<1,0,0>+<1,1,1,0>*<0,0,0,1>` (the bug reported).  This has been fixed by removing uses of `*Terms` as an input parameter to functions (but still using it as an output parameter).)

### DIFF
--- a/source/parser/parser.h
+++ b/source/parser/parser.h
@@ -755,7 +755,7 @@ class Parser : public SceneTask
         void Parse_Rel_Term (EXPRESS& Express, int *Terms);
 
         /// Parses a REL_TERM comparing two strings.
-        void Parse_Rel_String_Term (const UCS2 *lhs, EXPRESS& Express, int *Terms);
+        DBL Parse_Rel_String_Term (const UCS2 *lhs);
 
         /// Parses a LOGICAL_EXPRESSION (including FLOAT) or VECTOR.
         void Parse_Logical (EXPRESS& Express, int *Terms);
--- a/source/parser/parser.h
+++ b/source/parser/parser.h
@@ -755,7 +755,7 @@ class Parser : public SceneTask
         void Parse_Rel_Term (EXPRESS& Express, int *Terms);
 
         /// Parses a REL_TERM comparing two strings.
-        void Parse_Rel_String_Term (const UCS2 *lhs, EXPRESS& Express, int Terms);
+        void Parse_Rel_String_Term (const UCS2 *lhs, EXPRESS& Express, int *Terms);
 
         /// Parses a LOGICAL_EXPRESSION (including FLOAT) or VECTOR.
         void Parse_Logical (EXPRESS& Express, int *Terms);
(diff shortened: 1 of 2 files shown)


wfpokorny:
Hi Ben, 
Just looked this over and to me as an amateur c++ programmer your approach seems sound. 

I think this an important fix. It's one of the too long lived, nasty, little issues in POV-Ray I think causing users confusion.         

Questions: 

- Do you know if there was ever a FlySpray issue open against this issue? In searches now I did not turning anything up.  

- Also if anywhere other than the newsgroup we've previously documented that the precedence described at:  http://wiki.povray.org/content/Reference:Numeric_Expressions doesn't always hold? 

- Did you put together test scenes beyond those discussed in the newsgroup? If so, would you please add them as attachments here?   

I'm going to pull your req into my personal working version of POVRay and use it to help test. If you further tune your version of the code before something is merged into the master branch, please update your pull req so I can pick it up. 

Thanks.

source/parser/parser.h
```diff
@@ -755,7 +755,7 @@ class Parser : public SceneTask
         void Parse_Rel_Term (EXPRESS& Express, int *Terms);
 
         /// Parses a REL_TERM comparing two strings.
-        void Parse_Rel_String_Term (const UCS2 *lhs, EXPRESS& Express, int Terms);
+        void Parse_Rel_String_Term (const UCS2 *lhs, EXPRESS& Express, int *Terms);
```
c-lipka: From what I see, the new version of this function always sets `*Terms=1`, and only ever sets `Express[0]`; wouldn't it make more sense to give the function a full overhaul and change its signature to

    double Parse_Rel_String_Term (const UCS2 *lhs);

and leave all the worry about vector promotion up to the caller entirely?

source/parser/parser_expressions.cpp
```diff
@@ -1724,18 +1725,20 @@ void Parser::Parse_Rel_Factor (EXPRESS& Express,int *Terms)
 *
 ******************************************************************************/
 
-void Parser::Parse_Rel_String_Term (const UCS2 *lhs, EXPRESS& Express, int Terms)
+void Parser::Parse_Rel_String_Term (const UCS2 *lhs, EXPRESS& Express, int *Terms)
```
c-lipka: See above.

wfpokorny:
FYI. In my testing and other runs both the previous and current version of this fix are looking good thus far. Attaching a set of specific test cases for this too which cover what got posted to the newsgroup plus a few of my own. Otherwise have just been running scenes with this patch and thus far haven't been surprised. 

[BentSm_PromotePull_GH130.zip](https://github.com/POV-Ray/povray/files/602071/BentSm_PromotePull_GH130.zip)

PR #134, opened 2016-10-25
Flipping uv_mapping of torus in U to usual orientation.

wfpokorny:
See :

http://news.povray.org/povray.beta-test/thread/%3C580cd1a7%241%40news.povray.org%3E/

and attached scene file.

[mappings.pov.txt](https://github.com/POV-Ray/povray/files/550403/mappings.pov.txt)

If we decide to leave the torus uv_mapping as is closing this pull request, the documentation is currently vague enough we could just leave it to users to figure out this mapping different. Or we could document it is reversed in U from the usual rotation around Y for POV-Ray.

### DIFF
--- a/source/core/shape/torus.cpp
+++ b/source/core/shape/torus.cpp
@@ -1118,7 +1118,7 @@ void Torus::CalcUV(const Vector3d& IPoint, Vector2d& Result) const
     z = P[Z];
 
     // Determine its angle from the y-axis.
-    u = (1.0 - (atan2(z, x) + M_PI) / TWO_M_PI);
+    u = (atan2(-z,-x) + M_PI) / TWO_M_PI;
 
     len = sqrt(x * x + z * z);
 


c-lipka:
> If we decide to leave the torus uv_mapping as is closing this pull request, the documentation is currently vague enough we could just leave it to users to figure out this mapping different. Or we could document it is reversed in U from the usual rotation around Y for POV-Ray.

Unfortunately the behaviour is documented very precisely through its implementation in POV-Ray 3.7.0 (and, as I presume, earlier versions as well). Simply changing the mapping would potentially break existing scenes, and therefore isn't an option.

wfpokorny:
This is the documentation: 

torus : The map is taken from the area <0,0><1,1> where the u-coordinate is wrapped around the major radius and the the v-coordinate is wrapped around the minor radius. 

from http://wiki.povray.org/content/Reference:UV_Mapping 

Let me remind myself what Jerome said in the news group and look at what he might or might not have done in his hg-povray... His newsgroup response was open as to the torus change. And yep he did only what he said he'd do in hgpovray (the commit sha1 value being his): 

Jérôme Grimbert 0187cdd Correct Uv-mapping of cone, cylinder, lemon and ovus to be similar handness of sphere

I'm closing this out and will delete the branch.    

Lastly, I'll plan to include with the documentation to Jim for the new cone and cylinder mappings to add a note that the torus mapping is reversed from others in U.  I'll make  a post to the beta-test newsgroup this was what we decided too. 

Thanks.

PR #154, opened 2016-11-20
link with -lboost_date_time

rezso:
i686-uhu-linux-g++  -pipe -Wno-multichar -Wno-write-strings -fno-enforce-eh-specs -Wno-non-template-friend -s -minline-all-stringops -malign-double -pthread  -L/usr/lib  -L/usr/lib -o povray disp_sdl.o disp_text.o ../vfe/libvfe.a ../source/libpovray.a ../vfe/libvfe.a ../platform/libplatform.a -lSDL -lSDL -lpthread -lXpm  -lSM -lICE -lX11  -ltiff -ljpeg -lpng -lz -lrt -lm -lboost_thread  -pthread  -lboost_system
../source/libpovray.a(metadata.o): In function `boost::date_time::month_formatter<boost::gregorian::greg_month, boost::date_time::iso_extended_format<char>, char>::format_month(boost::gregorian::greg_month const&, std::ostream&)':
metadata.cpp:(.text._ZN5boost9date_time15month_formatterINS_9gregorian10greg_monthENS0_19iso_extended_formatIcEEcE12format_monthERKS3_RSo[_ZN5boost9date_time15month_formatterINS_9gregorian10greg_monthENS0_19iso_extended_formatIcEEcE12format_monthERKS3_RSo]+0x26): undefined reference to `boost::gregorian::greg_month::as_short_string()const'
metadata.cpp:(.text._ZN5boost9date_time15month_formatterINS_9gregorian10greg_monthENS0_19iso_extended_formatIcEEcE12format_monthERKS3_RSo[_ZN5boost9date_time15month_formatterINS_9gregorian10greg_monthENS0_19iso_extended_formatIcEEcE12format_monthERKS3_RSo]+0x48): undefined reference to `boost::gregorian::greg_month::as_long_string() const'
collect2: error: ld returned 1 exit status

### DIFF
--- a/unix/configure.ac
+++ b/unix/configure.ac
@@ -310,7 +310,7 @@ fi
 AC_MSG_CHECKING([whether the boost thread library is usable])
 SAVED_LIBS=$LIBS
 boost_thread_links=0
-for extralib in '' '-lboost_system'
+for extralib in '' '-lboost_system -lboost_date_time'
 do
   LIBS=$SAVED_LIBS
   LIBS="$LIBS $extralib"


c-lipka:
It should normally not be necessary to link the `boost_date_time` library (recently tested with boost 1.46 and 1.54); what version of boost are you using?

Also, while the proposed change may work around issues on some systems, it is less than ideal, as it unnecessarily forces linkage of `boost_date_time` on other systems, namely any that require linkage of `boost_system` but not `boost_date_time`.

Instead, if linkage of `boost_date_time` is indeed an issue, the `for extralib in ...` line should be changed to `for extralib in '' '-lboost_system' '-lboost_system -lboost_date_time'`, and the sample program (currently `boost::defer_lock_t(); return 0;`) changed to include whatever statement actually causes the requirement to link the `boost_date_time` library (my bet would be on `to_iso_extended_string()`.

c-lipka:
Also, can you please check whether `CXXFLAGS=-DBOOST_DATE_TIME_NO_LIB` solves the problem without linking the `boost_date_time` library? Thanks.

rezso:
I use boost 1.60 and gcc 6. The optimisation is disabled (--disable-optimiz).

rezso:
I tried now without the patch, and with CXXFLAGS=-DBOOST_DATE_TIME_NO_LIB.
I got linker error again, so the CXXFLAGS has no effect.

rezso:
Additional info: the patch needed only with --disable-optimiz. If I remove this option, the build is successful.

wfpokorny:
My non-optimized debug compiles too have needed LIBS="-lboost_date_time" for a while. I figured only developer types would be doing such compiles so I've just been adding the library when needed. 

Ubuntu probably back to 12.04 and various recent compilers and boost libraries.

c-lipka:
I'll add a corresponding note to the "Compatibility issues" section of `unix/install.txt`. Since `--disable-optimiz` is not the default, I think that's a cleaner solution than the one presented in this pull request.

PR #169, opened 2016-12-09
[ci skip] Updating the unix man page. Small typo.

wfpokorny:
Updating our unix man page to be current with recent 3.71 master.
Also small typo in processrenderoptions.cpp.

### DIFF
--- a/unix/povray.1
+++ b/unix/povray.1
@@ -1,14 +1,14 @@
-.TH POV-Ray 1 "August 2013" "POV\-Team" "Version 3.7" \" -*- nroff -*-
+.TH POV-Ray 1 "December 2016" "POV\-Team" "Version 3.71" \" -*- nroff -*-
 .\" man page written by Andreas Dilger
 .\" updated by Mark Gordon for POV-Ray 3.5
 .\" updated by Nicolas Calimet and Christoph Hormann for POV-Ray 3.6
 .\" updated by James Holsenback for POV-Ray 3.7
 
 .SH NAME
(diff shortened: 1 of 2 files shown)


unix/povray.1
```diff
@@ -1,49 +1,41 @@
-.TH POV-Ray 1 "August 2013" "POV\-Team" "Version 3.7" \" -*- nroff -*-
+.TH POV-Ray 1 "December 2016" "POV\-Team" "Version 3.71" \" -*- nroff -*-
```
c-lipka: Should probably still be `3.7` - if the version number was intended to refer specifically to 3.7.0, as opposed to the "3.7 generation" as a whole, I would expect it to have read `3.7.0` all along.

c-lipka:
How did you make the changes?

AFAIK the man pages used to be auto-generated from the HTML docs (though someone needs to dig up the script that does it, and run it).

unix/povray.1
wfpokorny: I was aiming for a man page which matched the 3.7.1 master branch if someone did a 'make install' after a build. I added -+CC added to the flags for example.  

I agree the previous should have been 3.7.0. 

Do you want something more generic? Only documentation correct for all 3.7.0 and 3.7.1 ?

wfpokorny:
I did straight up edits and the testing with the usual 'man -l povray.1' 

It looked from the authorship in the file as if it was just edited, but if you find that script, let me know and  I'll close this pull req. For the record if a script, we should change it to reformat the authors section as it looks bad on the default 80 char width unix terminal & there was a formatting typo - or two - in the result.

PR #176, opened 2016-12-19
[ci skip] Fixing copyright notices in scene dir html.

wfpokorny:
All others were already POV-Ray notices.

### DIFF
--- a/distribution/scenes/templates/Textures_Materials/Metals/__index.htm
+++ b/distribution/scenes/templates/Textures_Materials/Metals/__index.htm
@@ -359,9 +359,8 @@
 <hr size="3">
 <table width="95%" align="center">
 <tr>
- <td><i><font size="-1">&copy; Friedrich A. Lohm&uuml;ller, 2010<br>
-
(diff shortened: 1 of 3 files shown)


c-lipka:
Since this and pull requests #177 and #178 are all very fine-grained, while all being concerned with demo scene housekeeping, I'd suggest fusing them all together.

wfpokorny:
OK, I'll close all these scene related ones out and bundle them up. Probably next year before you see a larger package as I've got limited time for POV-Ray through year end.

c-lipka:
You might just as well create a single "demo scene housekeeping" pull request now, throw in everything you already have, and pile up more as you find the time. That way you could submit even minor changes whenever you feel like it, while I could squash-merge your work whenever if I feel like the time is ripe to get the latest and greatest, collapsing all your commits into a single one.

Squash-merging multiple pull requests is non-trivial and needs some serious Git knack, while squash-merging a single pull request with multiple commits is just a matter of a button click on GitHub.

PR #177, opened 2016-12-19
[ci skip] Adding torus2 example to scene dir html.

wfpokorny:


wfpokorny:
Closing to roll scene file updates into larger package.

Issue #180, opened 2017-01-07
Port of FS321 - Bounding threshold option handling.


wfpokorny:
http://bugs.povray.org/task/321

Also a part of what seen in issue #59.

---------------
Details:

User reported documentation inconsistency. Investigation led to the discovery of a bug in the setting of the current default value.

~source/frontend/renderfrontend.cpp reports the value "3" while ~source/backend/scene/scene.cpp sets a default value of "1"

Before for addressing this issue, are there any thoughts as to what the default value should be

---------------
Coments:

---
Comment by Jim Holsenback (jholsenback) - Friday, 21 March 2014, 16:44 GMT+5

change #6197 corrects inconsistency so that used value is the same as the reported value of "3"

wfpokorny:
Jim's fix, whatever it was, appears to have not made it to github but sounds like the value of 1 was changed to 3 in scene.cpp in perforce.   

I'll be submitting a pull request shortly which centralizes the default bounding threshold used in setting the threshold and reporting the threshold. 

Further, I've also restored the +MB-<n> clamping to 1+ which existed in 3.6 given the negative integers turn the bounding off and this is counter-intuitive given the value is a threshold value. 

Note some of the issues here come into play only when the user does NOT specify a bounding threshold value with bounding_threshold=<n> in an ini file or on the command line with +MBn.

Issue #59, opened 2016-04-27
Bounding issues related to +-MB and sphere_sweep's internal bounding.


wfpokorny:
(Other sphere_sweep issues: #147 #285 #287)

For further background see attached file and the thread at:

http://news.povray.org/povray.bugreports/thread/%3Cweb.5692847b8bcbbc2219f71b680%40news.povray.org%3E/
## My view as also posted to the above thread:

I was reminded of this issue the other day and have been doing some more investigation.

I spent a lot of time comparing 3.6 to 3.7 code given 3.6 works only to find nothing. I finally stumbled across the fact the bounding mechanism is always on in 3.7. In other words (-MB<n> / +MB<n>) doesn't function 3.7.

In 3.6 we need 3 shapes (I think) or more for the bounding mechanism to trigger and we have 1 so internal bounding is off.

Is this the -MB / +MB mechanism supposed to work in 3.7 ? We changed the default threshold in the docs to 25 so guessing yes. I could find no 3.6 -> 3.7 change notes related to it.

The issue behind is just another issue with the bounding computation as you thought Christoph, and it is sitting there in 3.6 too - if we turn bounding on with +MB1.

Basically the current mechanism multiplies the radius of each sphere by 2 and in this case here where the radius is small, but the range is a couple orders of magnitude larger, that 2x just isn't enough. This even though the curve is a pretty gentle cubic as curves go.

The multiplier should probably fudge upward from 2 by default if the range for the curve is large compared to the radius. Another more general option would be to add a new keyword to the sphere_sweep which lets the user optionally control the bounding radius multiplier?

[the.pov.txt](https://github.com/POV-Ray/povray/files/238516/the.pov.txt)

wfpokorny:
Just double checked the newsgroup link above and my last post is NOT there as I indicated though it is showing up in the 100 most recent posts thread. Not sure what happened in the web interface. The complete thread does show up properly in my thunderbird news reader.

c-lipka:
The post is there alright, currently being number 5 of 6 in the thread.

wfpokorny:
Yep, on manual reload the post is there. Firefox didn't pick up the updated page - should have thought to hit reload.

wfpokorny:
This issue tangled with github issue #147

wfpokorny:
A portion of the bounding issues here are related to issues with the bounding threshold control in 3.7. See issue #180 and the proposed pull request #181.

wfpokorny:
[...]

PR #181, opened 2017-01-07
Fixing issues with bounding threshold option.

wfpokorny:
Related to ussues #59 and #180. When no bounding threshold is specified
centralizes centralizes the setting and reporting of the bounding
threshold to 3. Further, restores the clamping of values to 1+ which
existed in 3.6.


c-lipka:
I think we should include this change into the current beta, as it seems straightforward enough, and besides it can well be classified as a bugfix.

If you agree, please rebase the pull request onto release/v3.7.1 (and change it to be merged into that branch), so I can pull it in there and propagate it to the master branch later with a straightforward merge. (I tried to effect the necessary changes from here, but that doesn't seem to work as intended.)

wfpokorny:
Agree. Give me a bit to figure out how to get this change onto a commit that lets me create the pull onto  release/v3.7.1. 

I'd already re-based this update onto a commit in the master branch later than the release/v3.7.1 branch commit.

wfpokorny:
Something just tried fails so I'm going to close this an open a new pull request to replace it.

PR #182, opened 2017-01-09
3.7 stable

Miskala:


c-lipka:
All those changes to the 3.7-stable branch are essentially changes already _in_ the master branch, and cherry-picked from there for the 3.7-stable branch. Merging them back into the master branch doesn't make sense.

If this pull request wasn't created by accident, please re-submit it, and provide a detailed description why we should want it in the master branch.

Miskala:
I apologize! It was an accident.

c-lipka:
No problem; one advantage of the pull request mechanism is that the impact of such accidents can be kept to a minimum.

(Also, even the official dev team members aren't immune to such accidents. Two such events have even left their idelible trace in the change history.)

Issue #183, opened 2017-01-09
Update scene files for best practice 3.7 onward gamma handling.


wfpokorny:
http://bugs.povray.org/task/106

(Issue originally opened way back in 2011 and looks to have never been completely implemented. I  argue strongly we not use conversion macros as suggested, but rather re-write where necessary for the cleanest presentation to _new_ users. Our example scenes should represent best current practices, not forever compatible, historical implementations. -WFP)

---------------
Details:

Most sample scenes and include files were designed at times when POV-Ray did not to any proper gamma handling, or still used the inferior 3.6 "assumed_gamma" mechanism.

All the scenes and include files should be reviewed, and updated to fit the new 3.7 gamma model.

The primary task will probably be gamma-adjusting literal color values and ambient parameters; I suggest using macros (which ideally should be defined in an include file) to be set according to the #version statement, so the scene/include file could be kept compatible with older versions.

c-lipka:
> All the scenes ***and include files*** should be reviewed, and updated to fit the new 3.7 gamma model.

Emphasis added because I think this is important - and also a problem, because existing scenes undoubtedly _do_ rely on the old behaviour.

I think the best timing for such an update would be a larger version step, e.g. 3.7.x to 3.8.0; and even then we would need to make sure we provide backward compatibility somehow.

c-lipka:
I think we really **want** to tackle this for v3.8.0, at least with respect to the scene files, and also the code snippets for the include menu. (The include files not so much, lest we break compatibility.) Therefore, I'm putting this on our plate for v3.8.0, to be tackled in the course of the beta phase.

I'm worried we might not find the time and energy though, so we may want to be open to reviewing this decision.

PR #185, opened 2017-01-16
Primarily updating pattern scene files for 3.7.1.

wfpokorny:
Including corresponding updates to preview html and jpgs.
Addresses issue #183 for the pattern examples.
Introduces aoi and potential pattern examples.


c-lipka:
Not sure what's going on there with the AppVeyor build, but I suspect it is related to branching from "master" but trying to merge into "release/v3.7.1".

distribution/scenes/textures/patterns/crackle_solid.pov
```diff
@@ -2,97 +2,190 @@
...
+#declare P_WoodGrain1B = pigment {
+    wood
+    scale <0.15,.5,1>
+    warp { turbulence <0.1,0.5,1> octaves 5 lambda 3.25 }
+    rotate <5,10,5>
+    translate -x*2
+}
+#declare T_Wood21 =
+    texture { pigment{ P_WoodGrain1A color_map { M_Wood11A }}}
+    texture { pigment{ P_WoodGrain1B color_map { M_Wood11B }}}
```
c-lipka: As for the colours include file, same goes for the woods include file.

distribution/scenes/textures/patterns/quilt2.pov
```diff
@@ -2,125 +2,130 @@
...
+//
 // -w800 -h600 +a0.3
 
 #version 3.7;
-
 global_settings { assumed_gamma 1.0 }
+#default { finish { ambient 0.006 diffuse 0.456 } }
 
-#include "colors.inc"
+#declare White = srgbft <1.00,1.00,1.00,0,0>;
```
c-lipka: Is there a reason to include filter and transmit into the white?

distribution/scenes/textures/patterns/aoi.pov
```diff
@@ -0,0 +1,100 @@
...
+// Persistence Of Vision Ray Tracer ('POV-Ray') sample file.
+//
+// Aoi pattern example aoi.pov.
+//
+// +w800 +h600 +a0.3
+
+#version 3.71;
+global_settings { assumed_gamma 1 }
+#default { finish { ambient 0.006 diffuse 0.456 } }
+#declare Black = srgbft <0,0,0,0,0>;
```
c-lipka: Is there a reason to include filter and transmit into the black?

distribution/scenes/textures/patterns/slope.pov
```diff
@@ -112,22 +112,29 @@ texture {
...
+#include "functions.inc"
+#declare Fn00 = function (x,y,z) {
+    pow(max(0,1-(x*x+y*y+(f_wrinkles(x,y,z)*2.5))),2.5)
+}
+#declare Fn01 = function (x,y,z) {
+    Fn00((x-0.5)*2,(y-0.5)*2,z)
+}
 height_field {
-   "slopemt_dat_36.jpg"
+   function 1200, 1200 { Fn01(x,y,z) }
```
c-lipka: Not sure if we want to replace the image-based height field with a function-based height field. Makes rendering of this particular scene easier, but probably isn't the standard use case for height fields.

distribution/scenes/templates/Textures_Materials/Metals/__index.htm
```diff
@@ -359,9 +359,8 @@
 <hr size="3">
 <table width="95%" align="center">
 <tr>
- <td><i><font size="-1">&copy; Friedrich A. Lohm&uuml;ller, 2010<br>
-
-   homepage:<a href="http://www.f-lohmueller.de"><nobr>http://www.f-lohmueller.de</nobr></font></i></a>
+ <td><i><font size="1">&copy; Persistence of Vision Ray Tracer, 2017<br>
```
wfpokorny: No, but the updated form is the copyright at the bottom in all but three of these html files from Friedrich. In looking at the pages nothing looked at all 'special' to me in a way which would make me think Friedrich wanted to maintain a copyright to just these 3 html pages.

distribution/scenes/textures.htm
```diff
@@ -464,7 +503,7 @@
 <hr size="3">
 <table width="95%" align="center">
 <tr>
- <td><i><font size="1">&copy; Persistence of Vision Ray Tracer, 2013<br>
+ <td><i><font size="1">&copy; Persistence of Vision Ray Tracer, 2017<br>
```
wfpokorny: Yes, I didn't think about this, but expect this is so...

distribution/scenes/textures/patterns/agate.pov
```diff
@@ -2,35 +2,38 @@
...
 // California, 94041, USA.
-
-// Persistence Of Vision raytracer sample file.
-// Agate pattern example
 //
-// -w320 -h240
+// Persistence Of Vision Ray Tracer ('POV-Ray') sample file.
+//
+// Agate pattern example agate.pov.
+//
```
wfpokorny: It was used inconsistently - sometimes also wrong in aspect ratio for the scene (though maybe not in this batch of updates...) - and these days expect the smaller resolution of little use being there I expect for reasonable performance.

distribution/scenes/textures/patterns/agate.pov
```diff
@@ -2,35 +2,38 @@
...
 // -w800 -h600 +a0.3
 
 #version 3.7;
-
 global_settings { assumed_gamma 1.0 }
+#default { finish { ambient 0.006 diffuse 0.456 } }
```
wfpokorny: If we are willing to go with less matched to 3.6 results  these are options. 

FYI - I did compares to 3.6.1 for most scenes changed - certainly all the more complex ones.

distribution/scenes/textures/patterns/agate.pov
```diff
@@ -2,35 +2,38 @@
...
 
-#include "colors.inc"
+#declare Black = srgb <0,0,0>;
+#declare White = srgb <1,1,1>;
```
wfpokorny: Understand. Done should new users grab one of these scenes and playing, tweak colors to values they might grab or know from the web. In other words the 'srgb' is there so the the scene is more likely to work as they expect.

distribution/scenes/textures/patterns/agate.pov
```diff
@@ -2,35 +2,38 @@
...
+#declare T1 = texture {
+    pigment {
+        agate
+        color_map {
+            [0.0 Black]
+            [1.0 White]
+        }
+        scale 0.5
+    }
+}
```
wfpokorny: Not opposed to further change. I picked a standard close to our C++ code base for indentation & what I use personally as a standard for these updates given there were previously many in this directory - let alone across all the scene and include files. 

Also something which didn't cram large amounts of 'function' onto one line - again thinking of users new to POV-Ray. I don't feel 'short' code should be our aim when instructing.

distribution/scenes/textures/patterns/crackle_solid.pov
```diff
@@ -2,97 +2,190 @@
...
 
 //--textures definitions--------------------------------
-
+#declare Black  = srgb <0,0,0>;
+#declare Gray60 = srgb <0.6,0.6,0.6>;
+#declare Gray80 = srgb <0.8,0.8,0.8>;
+#declare Gray90 = srgb <0.9,0.9,0.9>;
+#declare Red    = srgb <1,0,0>;
+#declare Tan    = srgb <0.86646,0.58571,0.44532>;
+#declare Wheat  = srgb <0.85494,0.85494,0.75836>;
```
wfpokorny: Perhaps some of this is my experience. Remembering back to my start with POV-Ray I found the includes for color.inc NOT useful. The colors beyond a very few did not correspond to anything I used. Today I look at these and wonder their source - not to mention the SDL encoding is odd when new users go looking for colors.inc. 

I know you encouraged the colors.inc overhaul & I looked at it and other includes, but dang this ripples into everything. I'm just not ready to dive into that amount of grunt work and inevitable inevitable complaints some scene or other doesn't look right. 

In one scene's changes (tiling.pov?) I made use of SVG1.0 standard colors which are almost identical to X11 and .net standards for the web. I have the full set encoded in my working space & have been playing with it. My thinking is maybe we decouple the old POV-Ray colors in colors.inc from anything 3.7 and newer by going with the SVG1.0 set for anything updated, but not convinced of this path myself yet. 

What I decided to do for now is update and decouple from 'gamma' tangled includes for now. We can more quickly get to a set of scenes for new users that are not 'washed out' in appearance.  Such a  'step' doesn't preclude further updates.

distribution/scenes/textures/patterns/crackle_solid.pov
wfpokorny: Please see my comment to the colors.inc decoupling for explicit srgb encoding for the bulk of my reasoning. 

I'll add for all the woods.inc we have the issue these overlap textures with transparency - which as I highlighted in a gamma related post tp the newsgroup doesn't work as 'tightly' in an assumed_gamma '1' space. I implemented something for assumed_gamma 1, but I don't like it as an equivalent replacement. The local coding is better, but a match requires more than the map I did.

distribution/scenes/textures/patterns/crackle2.pov
```diff
@@ -2,207 +2,189 @@
...
+      #local T_CrackColor00 = srgb <0.52157,0.40000,0.72549>;
+      #local T_CrackColor01 = srgb <0.83137,0.66667,0.92157>;
+      #local T_CrackColor02 = srgb <0.83137,0.66667,0.92157>;
+      #local T_CrackColor03 = srgb <0.87059,0.66667,0.92941>;
+      #local T_CrackColor04 = srgb <0.95686,0.75294,0.95294>;
+      #local T_CrackColor05 = srgb <1.00000,0.84314,0.93333>;
+      #local T_CrackColor06 = srgb <0.98823,0.85490,0.79216>;
+      #local T_CrackColor07 = srgb <0.97647,0.86667,0.67059>;
+      #local T_CrackColor08 = srgb <0.98823,0.86667,0.60784>;
+      #local T_CrackColor09 = srgb <0.58431,0.48628,0.90588>;
```
wfpokorny: Somewhat agree about the disconnect here, but there are reasons. 

1) This is the form I use and others on the newsgroups have adopted so the multiplication of the color - often done as a quick tweak - actually works as the user expects. This scene doesn't have it but others do multiplication in the color_map from defined colors like this to show the form which works.  

2) Normally the color names would mean something is this form as in SVG_redorange, but in trying to match 3.6 results as closely as possible I was stuck with either going with generic names like those above or coming up with names that might be close to some web color people might know. 

3) Not true here, but very often colors are repeated in color maps and a common definition makes this much more clear.

distribution/scenes/textures/patterns/quilt2.pov
wfpokorny: It is my habit to code the full 5, but the above is a mistake. I meant to use srgb - or the shortest srgb* form for each situation. I'll push an update for this one and others I find.

distribution/scenes/textures/patterns/aoi.pov
wfpokorny: See previous comment.

distribution/scenes/textures/patterns/slope.pov
wfpokorny: 1) We have the crater (speling?) example with an image. 

2) Here the previous creation method involved overlapping textures - non zero transmit. As I mentioned above and in the news groups - the transmit mechanism doesn't work as tightly in the assumed_gamma 1.0 space. I didn't want to deal with trying to get something to match, or recommend the image creation be done in the 2.2 space. 

3) The set up to create the image is out of date & I thought a bit odd - and so not the best example - in not using an orthographic camera set up when creating the image for the mountain.

wfpokorny:
A general comment. These scene updates, like our documentation updates, are grunt work. In my case I feel an obligation to give back to this thing called POV-Ray which as my hobby has done much to keep me glued together over the years.

I share - and appreciate - the desire for perfection in what we do here. 

However, the documentation, our sample scenes (less the include files), and the like are much more amenable to gradual improvement. Of course outright errors and bugs should be fixed, but I'm suggesting the measure for support aspects be more like: "Does this update make the POV-Ray package substantially better?"  Nothing prevents further refinement.   

There is a limit on how much of my hobby time I am willing to burn on churn and grunt work.

distribution/scenes/templates/Textures_Materials/Metals/__index.htm
c-lipka: You have a point there that I didn't notice, and I suspect you're right. But when it comes to legal stuff, I think we should be cautious, and at least check with Chris, if not with Friedrich.

distribution/scenes/textures/patterns/agate.pov
c-lipka: I thought we might keep it around for thumbnails. Of course they should be consistent in aspect ratio with the larger image.

distribution/scenes/textures/patterns/agate.pov
c-lipka: As I'm saying, we can't perfectly match the 3.6 results, even when it comes to "simple" stuff like ambient and diffuse: As soon as they combine with any other effect (or even with each other), the simple power-law relationship comes crashing down. So rather than use precisely calculated values that only match under hypothetical circumstances, I'd advocate nicely rounded values.

distribution/scenes/textures/patterns/agate.pov
c-lipka: I'm not arguing against the use for "srgb", but throwing out "colors.inc".

distribution/scenes/textures/patterns/agate.pov
c-lipka: I understand and fully go along with your motivation. But as I said, if we go for unification of the style, I think we should first discuss what to actually go for. Not the least because such a unification would also touch areas beyond sample scenes, such as documentation (embedded samples) and even the POV-Ray source code (e.g. editor default settings like tab width, which I think I dimly recall as having the ghastly value of 3 in a fresh install).

distribution/scenes/templates/Textures_Materials/Metals/__index.htm
wfpokorny: OK. Let me post something to the dev group about this. 

Do you know if Friedrich on copy to such notes?

distribution/scenes/textures/patterns/agate.pov
wfpokorny: If they are used to create thumbnails, I guess we can put them in everywhere? Do we know this resolution is used? The html thumbnails are mostly 160 by 120 - except when they are not and then I tried to keep the height constant. Suppose some similar approach could be taken using 240 as a fixed height.

distribution/scenes/textures/patterns/agate.pov
wfpokorny: Alright. I'll update to 0.01 and 0.5 for the ambient and diffuse '#default' and add albedo. For the record, the last is more likely to cause larger changes when it comes into play in my experience.

distribution/scenes/textures/patterns/agate.pov
wfpokorny: OK. I'm not signing up for the colors.inc update in the near term. If you really want this, I suggest we close this request and cherry pick the introduced examples so we at least have those.

distribution/scenes/textures/patterns/agate.pov
wfpokorny: Sounding more like we should close this pull...

distribution/scenes/templates/Textures_Materials/Metals/__index.htm
c-lipka: Dunno; I'd have to ask Chris.

distribution/scenes/textures/patterns/agate.pov
c-lipka: As I'm saying, there's no need to make any such decision in _this_ case.

As for the _general_ question of whether to cut ties with `colors.inc` where necessary to improve the sample scenes, or live with sub-standard sample scenes as long as we have sub-standard include files (which might be the more honest thing to do), I'm torn apart myself. I'd really love to get more feedback from other people on this one before deciding either way.

distribution/scenes/textures/patterns/crackle_solid.pov
c-lipka: Maybe a standard like SVG or HTML colour names would provide an easy way out of the conundrum: If we provide a file "colors_web.inc" or some such, we can recommend that one for new scenes, use it in sample scenes, and deprecate the old "colors.inc".

Do you happen to know where SVG and HTML colours differ? (X11 colours are apparently not an option, since they are defined without reference to a particular colour space; SVG and HTML on the other hand are defined in terms of sRGB colours.)

wfpokorny:
FYI. I'm going to hold off any further updates & follow up despite some comments to the contrary above. Given the reasonable wishes as to direction, the whole of this update isn't going to fly anything close to as is. I'm closing it.

Tomorrow, I'll pull out two smaller updates from this mess as new pull requests. Namely, the new aoi and potential demo scenes.    

I saw the developer note related to this req earlier; Don't read any exterior meaning into my bailing here. The decision comes to me not being interested in investing additional time in scene and include updates - maybe someday. I've already other 'work' to finish in porting or handling the remaining open flyspray issues. I'm otherwise going to get back to ideas on my POV-Ray 'play' list which includes some interest in fixing bugs.

PR #231, opened 2017-02-05
Adding POV_THREAD_STACK_SIZE macro.

wfpokorny:
Thread stack size defaulting refined. New adjustments for
64 bit compiles vs 32 bit and linux needing a larger stack size
than windows for one recent pre 371 regression.

New sub macros POV_COMPILE64 and POV_THREAD_STACK_SIZE_DEFAULTED
created in support of POV_THREAD_STACK_SIZE macro. See Doxygen
comments in code for more detail.

See newsgroup thread: 
http://news.povray.org/povray.beta-test/message/%3C589458a6%40news.povray.org%3E/#%3C589458a6%40news.povray.org%3E

A small hmm.cpp file attached which was used for test the macros and various overrides on my Ubuntu machine:  
[hmm.cpp.txt](https://github.com/POV-Ray/povray/files/753195/hmm.cpp.txt)

### DIFF
--- a/source/backend/support/task.cpp
+++ b/source/backend/support/task.cpp
@@ -8,7 +8,7 @@
 /// @parblock
 ///
 /// Persistence of Vision Ray Tracer ('POV-Ray') version 3.7.
-/// Copyright 1991-2016 Persistence of Vision Raytracer Pty. Ltd.
+/// Copyright 1991-2017 Persistence of Vision Raytracer Pty. Ltd.
 ///
 /// POV-Ray is free software: you can redistribute it and/or modify
 /// it under the terms of the GNU Affero General Public License as
@@ -103,7 +103,7 @@ POV_LONG Task::ConsumedCPUTime() const
 void Task::Start(const boost::function0<void>& completion)
 {
     if((done == false) && (taskThread == NULL))
-        taskThread = NewBoostThread(boost::bind(&Task::TaskThread, this, completion), 1024 * 1024 * 2); // TODO - make stack size definable
+        taskThread = NewBoostThread(boost::bind(&Task::TaskThread, this, completion), POV_THREAD_STACK_SIZE);
 }
 
 void Task::RequestStop()
--- a/unix/povconfig/syspovconfigbackend.h
+++ b/unix/povconfig/syspovconfigbackend.h
@@ -11,7 +11,7 @@
 /// @parblock
 ///
 /// Persistence of Vision Ray Tracer ('POV-Ray') version 3.7.
-/// Copyright 1991-2016 Persistence of Vision Raytracer Pty. Ltd.
+/// Copyright 1991-2017 Persistence of Vision Raytracer Pty. Ltd.
 ///
 /// POV-Ray is free software: you can redistribute it and/or modify
 /// it under the terms of the GNU Affero General Public License as
@@ -49,5 +49,18 @@
 // On Unix platforms, we don't do anything special at thread startup.
 #define POV_USE_DEFAULT_TASK_INITIALIZE 1
 #define POV_USE_DEFAULT_TASK_CLEANUP    1
+
+/// @note
+///     In regression testing an old scene for the 3.7.1 release found
+///     linux machines needed more stack storage than windows.
+///     See too @ref POV_THREAD_STACK_SIZE
+///
+#ifndef POV_THREAD_STACK_SIZE
+    #define POV_THREAD_STACK_SIZE (1024 * 1024 * 4)
+#else
+    #if POV_THREAD_STACK_SIZE < (1024 * 64)
+        #error "POV_THREAD_STACK_SIZE set less than 65KB or not a byte count."
+    #endif
+#endif
 
 #endif // POVRAY_UNIX_SYSPOVCONFIGBACKEND_H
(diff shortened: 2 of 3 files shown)


c-lipka:
I noticed that you're using a lot of conditionals to set the stack size depending on the platform. I would highly recommend to use the most platform-specific config files for that instead. The canonical mechanism is as follows:

- `source/MODULE/configMODULE.h` should define a very generic last-ditch resort default (wherever possible/applicable) for any settings primarily relevant for the module, as well as include Doxygen-style documentation for the setting, using the `#ifndef MACRO #define MACRO ... #endif` idiom.

- `INCARNATION/povconfig/syspovconfigMODULE.h` and `INCARNATION/povconfig/syspovconfig.h` are effectively pulled in earlier, and may define settings specific to an incarnation (currently "unix" = POV-Ray for Unix, or "windows" = POV-Ray for Windows). The former is pulled in only for the specific module (and all modules depending on it), while the latter is pulled in even earlier for all modules. In general, these files should be used _only_ for stuff common to all variants of the incarnation (e.g. independent of build tool used, specific OS variant, CPU architecture etc.)

- `INCARNATION/povconfig/syspovconfig_VARIANT.h` is effectively pulled in first, and is intended to define any settings not common to all variants (e.g. stuff specific to a particular Version of MSVC, to 32 vs. 64 bit, or any some such).

Using this mechanism may also make it unnecessary to have two different `POV_THREAD_STACK_SIZE[_*]` macros, which I find odd (I haven't had a close enough look to understand what you are actually doing there), and will most certainly avoid the need for a `POV_COMPILE64` macro.

source/backend/support/task.cpp
```diff
@@ -103,7 +103,7 @@ POV_LONG Task::ConsumedCPUTime() const
 void Task::Start(const boost::function0<void>& completion)
 {
     if((done == false) && (taskThread == NULL))
-        taskThread = NewBoostThread(boost::bind(&Task::TaskThread, this, completion), 1024 * 1024 * 2); // TODO - make stack size definable
+        taskThread = NewBoostThread(boost::bind(&Task::TaskThread, this, completion), POV_THREAD_STACK_SIZE);
```
c-lipka: We might design the mechanism so that the boost standard stack size is used if POV_THREAD_STACK_SIZE is undefined.

c-lipka: We might design the mechanism so that the boost standard stack size is used if `POV_THREAD_STACK_SIZE` is undefined.

wfpokorny: What I found was boost doesn't have a standard stack size on linux & likely any platform. It inherits whatever the parent process stack setting is and this exposes us to the stack size being too small (or just different to our own) due a particular user's environment. 

Today anyone post Jan 16' and >Boost 1.5 gets a 2MB hard coded stack. I believe we should stick with this hard coded minimum values approach on all platforms given the opaque fail signature and that we know existing user scenes need this amount or more.

"POV_THREAD_STACK_SIZE" therefore needs to be defined for all platforms. The macro itself apparently in part due Doxygen. If we use it directly, want to adjust for 64 bits on as many platforms as possible, adjust again for linux and still be able to detect whether the user has set his own value so we don't override it - things get somewhat complicated... As for following the form for the headers files, I suppose I think I reasonably did. :-)  

Anyway, I'm not going to burn a lot more time here. Give me a bit to see if I can get something simpler working which tactically fixes the known regression fail on linux and still lets us set POV_THREAD_STACK_SIZE. If I can, I'll force an update here later this morning, otherwise I'll close this req.

wfpokorny:
OK. A simpler implementation pushed to this request which rips out the 64bit vs 32bit stuff and corrects for me getting tangled up on the header inclusion order. Thanks for the earlier prompt on the subject :-).

source/backend/support/task.cpp
c-lipka: We still have a bit of time before beta.3, as the AVX stuff still needs some work as well, so there's no reason for excessive haste.

If you find you can't improve things beyond the current point, please do _not_ close the pull request; I think it's better than the status quo, so I guess I'd like to pull it either way, and maybe amend it later if I find the time.

As a general rule, if I just use regular comments on a pull request, consider them mere suggestions for further improvement; If I should find something that I consider a no-go, I would explicitly flag it as "change required" (or some such; can't tell off the top of my head what GitHub's official term for that review mechanism is).

unix/povconfig/syspovconfigbackend.h
```diff
@@ -50,4 +50,17 @@
 #define POV_USE_DEFAULT_TASK_INITIALIZE 1
 #define POV_USE_DEFAULT_TASK_CLEANUP    1
 
+/// @note
+///     In regression testing an old scene for the 3.7.1 release found
+///     linux machines needed more stack storage than windows.
+///     See too @ref POV_THREAD_STACK_SIZE
+///
```
c-lipka: Doxygen may fail to recognize this comment to be associated with `POV_THREAD_STACK_SIZE` if the macro is already set earlier; hence the `/// @def POV_FOO` whenever we provide defaults for a macro in `source/base/configbase.h` and similar places, which explicitly instructs Doxygen that the comment block applies to that macro.

My suggestion would be to make this a plain old C++ comment for now, since we're not generating Doxygen docs from the platform-specific configs at the moment anyway.

Issue #48, opened 2015-11-15
Degraded Performance


c-lipka:
Performance has declined since the official 3.7.0 release.

On Ubuntu 14.04.04, the inbuilt benchmark is reported to be running about 19% slower for 3.7.1-alpha.8344733; on Windows 7, a difference of about 17% has been measured.

The reason for this difference needs to be examined.

toddcarnes:
I ran some unofficial tests of my own as follows:

1. I downloaded and installed the official POV-Ray for Windows 3.7.0.0.
2. I ran the official benchmark test 3 times then averaged the results.
3. I downloaded & compiled from git POV-Ray for Windows 3.7.1-alpha.8344735.unofficial.msvc14.win64.
4. I ran the official benchmark three times and averaged the results.

My results basically agreed with yours. Official version 3.7.0.0 runs 1.23 times faster than unofficial version 3.7.1.alpha. However, I also noticed another significant difference. Official version 3.7.0.0 also uses 1.83 times as much memory as version 3.7.1.alpha.

So, it would appear that you have traded a slower rendering speed for better use of available memory.

A more detailed version of my testing is in the attached file.
[povrayTR.txt](https://github.com/POV-Ray/povray/files/35118/povrayTR.txt)

wfpokorny:
I had occasion to compile both stable and master today so checked again the performance of the benchmark as I know master has changed quite a lot since this was opened. 

Ubuntu 14.04
stable -> 3.7.1-alpha.8581629
CPU    1104.82 seconds -> 1310.64 seconds 18.63%
Memory 34824KB -> 34296KB -1.52%

As before reported statistic more or less the same - media samples up 1.52% .  I'll put it on my list to later attempt a couple profiled runs.

wfpokorny:
Attached is a quick view of gprof results comparing 3.7.0-stable to 3.7.1-alpha.8581629.

Both compiled with all optimization off and run with one thread with the command: 
povray --benchmark +wt1

The full output is about 30MB - most of it related to things not consuming much time.

[Quick_gprof_view.txt](https://github.com/POV-Ray/povray/files/238504/Quick_gprof_view.txt)

wfpokorny:
In looking to clean up some old FlySpray tasks had occasion to run the attached SplineThreadSafety.pov file in both 3.7.0-Stable and our 3.7.1 branch at commit : 598f29aff140675b99305fdcb051482bb5ed7bbd 

[...]

PR #258, opened 2017-03-24
Recovering about 25% of benchmark slowdown due commit 725efe1.

wfpokorny:
Specifically, adding warp type tracking to the warp constructors as a
way to use static_cast over dynamic_cast calls at run time.

Patterns with the most potential benefit are: Agate, Marble, Spiral1,
Spiral2 and Wood.

See github issue #48 for more detail.

### DIFF
--- a/source/core/material/pattern.cpp
+++ b/source/core/material/pattern.cpp
@@ -14,7 +14,7 @@
 /// ----------------------------------------------------------------------------
 ///
 /// Persistence of Vision Ray Tracer ('POV-Ray') version 3.7.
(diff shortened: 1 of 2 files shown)


c-lipka:
In terms of design this change would be a step backward, so I'll be going for a different approach to eliminate the `dynamic_cast` from `SearchForTurb()`.

Thanks for identifying it as another contributor to the performance loss though. I guess I better review all other `dynamic_cast`s as well, and eliminate any others that might get invoked as part of the actual ray tracing.

PR #267, opened 2017-04-06
Add AVXPortable and changed noise code selection

anshuarya:
1) Added new noise implementation "AVXPortable" which is simply PortableNoise compiled separately with compiler AVX flags.

2) Changed the noise code selection based on CPU vendor because the existing AVX and AVX2FMA3 implementations lower performance on non-Intel CPUs compared to the new AVXPortable or existing portable (non-AVX) noise functions.


c-lipka:
Thanks a lot for your contribution; I'll pull it as soon as I find the time (presumably start of the week).

PR #270, opened 2017-04-10
Improving AA performance for non MSVC compilers.

wfpokorny:
Web searches turned up evidence the existing code has the best
performance for Microsoft Visual C++, but performance is awful for g++
and clang compared to sampled.assign method.

g++ (5.4.0) +a0.05 +am2 +r5 +w1800 +H1200 complete render
(threads==cores) is 23% slower with iterator method.

clang (3.8.0-2) with AA above is 28% slower using iterator.

### DIFF
--- a/source/backend/render/tracetask.cpp
+++ b/source/backend/render/tracetask.cpp
@@ -8,7 +8,7 @@
 /// @parblock
 ///
 /// Persistence of Vision Ray Tracer ('POV-Ray') version 3.7.
-/// Copyright 1991-2016 Persistence of Vision Raytracer Pty. Ltd.
+/// Copyright 1991-2017 Persistence of Vision Raytracer Pty. Ltd.
 ///
 /// POV-Ray is free software: you can redistribute it and/or modify
 /// it under the terms of the GNU Affero General Public License as
@@ -216,8 +216,7 @@ RGBTColour& TraceTask::SubdivisionBuffer::operator()(size_t x, size_t y)
 
 void TraceTask::SubdivisionBuffer::Clear()
 {
-    for(vector<bool>::iterator i(sampled.begin()); i != sampled.end(); i++)
-        *i = false;
+    sampled.assign(sampled.size(), false);
 }
 
 TraceTask::TraceTask(ViewData *vd, unsigned int tm, DBL js, DBL aat, unsigned int aad, pov_base::GammaCurvePtr& aag, unsigned int ps, bool psc, bool final, bool hr) :


c-lipka:
> g++ (5.4.0) +a0.05 +am2 +r5 +w1800 +H1200 complete render (threads==cores) is 23% slower with iterator method.

What scene would that be?

My assumption would be that the impact of SubdivisionBuffer::Clear() on render performance would diminish for any non-trivial scene, and therefore not warrant maintaining different platform-specific implementations. So I would suggest examining which build environment exhibits the biggest difference in performance between the two implementations, and go for whatever runs fastest there.

_If_ we go for platform-specific implementations, then we shouldn't implement them _in situ_, but rather declare a generic function (e.g. some "FillBoolVector(std::vector<bool>, bool)"), and implement it in the platform-specific section of the source code. That way we could re-use that implementation for any other bool vectors we might be using.

wfpokorny:
My test scene was 4 approximately 20 character text objects with a plain color texture taking up maybe 10% of the pixels. I ran some other scenes, but they were all relatively simple in structure.   

I'm running the shipped bencmark.pov. benchmark.ini now except with +a0.05 +am2 +r5. On my little I3 it's going to take a while to turn both before and after at +wt2 - which I use to avoid volatility in results due cache use, register starving when threads > cores. Media+turb/noise1 taking 70-80% of the time with this scene normally will tend to wash out the AA performance difference I expect.  

That said, I think this non-MSVC effect is so large, it will be there with any scene. Even the scenes I ran at +r3 were always at least 5% slower. Anyway, I'll report back when I have the benchmark results. 

Remember: I don't have a windows machine (MSVC compiler) on which to test any common fix, which is the reason I didn't go for the common approach in the code after finding web results showing the iterator best with the Microsoft compiler.

c-lipka:
My own experiments indicate that the information you have re the MSVC compiler may be outdated; with VS 2015, I see better performance with the `assign` approach than with the naive loop.

Running "Basic Scene 01" from the insert menu with `+w8000 +h6000 +a0.05 +am2 +r5 +wt1`, I get the following performance values on my i7-5820 (pixels per second of CPU time):

- 164624.42 for the naive loop
- 176846.47 for `sampled.assign(sampled.size(), false)`
- 176714.22 for `sampled.clear()` followed by `sampled.resize(...)`
- 184007.19 for `std::fill(sampled.begin(), sampled.end(), false)`

Since we're no longer supporting earlier versions of MSVC, we can therefore safely ditch the naive loop in favour of a faster approach. I'd oppose the `clear`/`resize`-based one because I consider it ugly and potentially problematic, but both your proposal as well as the `fill`-based implementation look fine to me.

According to <http://stackoverflow.com/questions/19562667/how-to-zero-a-vectorbool>, g++ should provide the best performance with the `fill`-based implementation, too, but that information might be outdated as well, and clang may behave still differently.

wfpokorny:
Thanks much Christoph for the MSVC information (the pixels per sec method a useful short cut for performance comparisions!). The short of it is " sampled.assign(sampled.size(), false);" looks like the best choice and I'll push such an update to this pull request in a bit. 

You were right the biggest benefit comes with trivial / less complex scenes, but everything I ran shows some gains in performance. I also scanned the for similar vector<bool>::iterator elsewhere in the code, but turned up nothing. 

Here is a summary of some of what I ran yesterday for the record: 

Using potential.pov from the scenes collection
-------------------------------------------------------------

for(vector<bool>::iterator i(sampled.begin()); i != sampled.end(); i++)
    *i = false;
--- (clang) +a0.05 +am2 +r5 +W1200 +H1200 +wt2 -D
10.71user 0.05system 0:06.33elapsed
10.73user 0.03system 0:06.33elapsed
--- (g++) +a0.05 +am2 +r5 +W1200 +H1200 +wt2 -D
10.14user 0.04system 0:06.05elapsed
10.19user 0.02system 0:06.05elapsed

sampled.assign(sampled.size(), false);
--- (clang) +a0.05 +am2 +r5 +W1200 +H1200 +wt2 -D
8.06user 0.04system 0:05.02elapsed   -24.95%
8.03user 0.05system 0:05.02elapsed
--- (g++) +a0.05 +am2 +r5 +W1200 +H1200 +wt2 -D
7.96user 0.04system 0:04.96elapsed   -22.13%
7.85user 0.04system 0:04.91elapsed

std::fill(sampled.begin(), sampled.end(), false);
--- (clang) +a0.05 +am2 +r5 +W1200 +H1200 +wt2 -D
8.21user 0.02system 0:05.08elapsed   -23.60%
8.17user 0.04system 0:05.06elapsed
--- (g++) +a0.05 +am2 +r5 +W1200 +H1200 +wt2 -D
8.08user 0.04system 0:05.02elapsed   -20.91%
8.00user 0.03system 0:04.96elapsed

Other comparisons
-------------------------------------------------------------

---- g++ (+wt4) benchmark.ini +a0.05 +am2 +r5
iterator       33427.87user 5.08system 2:20:37elapsed
sampled.assign 33348.85user 4.34system 2:19:58elapsed -0.24% (faster)

---- g++ (+wt4,+wt2) balcony scene +a0.05 +am2 +r5
iterator       4807.56user 1.02system 20:20.12elapsed
sampled.assign 4788.52user 1.01system 20:13.29elapsed -0.40% -0.31%(?)

---- g++ (+wt4) text object 80 chars +a0.05 +am2 +r5
iterator       11.63user 0.02system 0:03.89elapsed
sampled.assign 10.67user 0.02system 0:03.68elapsed -8.25%

---- g++ (+wt2) text object 80 chars +a0.05 +am2 +r5
iterator       7.50user 0.04system 0:04.44elapsed
sampled.assign 6.71user 0.02system 0:04.08elapsed -10.53%


Less agressive AA, less benefit (potential.pov)
-------------------------------------------------------------

--- (clang) +a0.1 +am2 +r3 +W1200 +H1200 +wt2 -D
for(vector<bool>::iterator i(sampled.begin()); i != sampled.end(); i++)
    *i = false;

4.51user 0.02system 0:03.23elapsed  
4.50user 0.02system 0:03.23elapsed

--- (clang) +a0.1 +am2 +r3 +W1200 +H1200 +wt2 -D
sampled.assign(sampled.size(), false);

4.23user 0.03system 0:03.08elapsed   -5.99%
4.24user 0.01system 0:03.11elapsed

Issue #65, opened 2016-05-18
Normal for the prism's bezier spline mode is reversed from other spline types & lathe uses.


wfpokorny:
The issue is that the normal for the prism bezier spline is reversed from all other spline uses in both lathes and prisms. See the attached tarball for the README, test cases, some pre-rendered.

[PrismBezierNormalFlipped.tar.gz](https://github.com/POV-Ray/povray/files/271192/PrismBezierNormalFlipped.tar.gz)

wfpokorny:
New version of test cases where both prism ends are visible to the camera given the discovery here: 

http://news.povray.org/57533e83%40news.povray.org

[PrismBezierNormalFlipped_V0.1.tar.gz](https://github.com/POV-Ray/povray/files/299339/PrismBezierNormalFlipped_V0.1.tar.gz)

LeForgeron:
There is two issues, separate:
* the orientation of bezier_spline prism's side normal
* for all prism, the lower face has the same orientation as the upper face, so interior_texture is doomed to fail.

wfpokorny:
Let us add a third normals issue to this github issue...

* The normals for sides of text objects is also flipped from what it should be. In searches found this has been mentioned on and off in the news groups. 

In the attached image we should not see the red interior texture at all, but we do on the text object sides. 

![textobjectsidenormalflipped](https://cloud.githubusercontent.com/assets/15925553/24864115/2f7dd0be-1dd1-11e7-8031-9e3e7a95d6a6.jpg)

LeForgeron:
For the text object, It might be fixed with 

https://github.com/LeForgeron/povray/commit/2595932240ea8caa99b2a9c610c8af9d90ff0558

(yep, it is !)

wfpokorny:
Thanks  Jérôme. Now that you mention it I do remember your version of POV-Ray getting this fix...  I did it in my pull request as an afterthought to the performance changes which where my primary interest.

wfpokorny:
Closing. 

Ran all the tests again plus additional text cases to look at the normal changes there and all is fixed with e7aac79. 

For anyone picking up these tests again note a few stacked ring orthographic camera ones are noisy at the edges. Further there have been change from say v370 to v380 - the ambient default for example which affects the test case results herein.

PR #271, opened 2017-04-10
Fixing text object side normals. Improving performance.

wfpokorny:
Varies with font, but with longer strings text object performance
5-20% better. Where using inside object tests specifically and
heavily have seen as much as 2-3x better performance.

Some general code clean up too. Eliminated all but one goto statement
in truetype.cpp. Re-worked very old code which looked to be optimized
for CPUs without floating point capability.

This pull requests addressed the text object normal flip of issue #65.

### DIFF
--- a/source/core/shape/truetype.cpp
+++ b/source/core/shape/truetype.cpp
@@ -10,7 +10,7 @@
 /// @parblock
 ///
 /// Persistence of Vision Ray Tracer ('POV-Ray') version 3.7.
-/// Copyright 1991-2016 Persistence of Vision Raytracer Pty. Ltd.
+/// Copyright 1991-2017 Persistence of Vision Raytracer Pty. Ltd.
 ///
 /// POV-Ray is free software: you can redistribute it and/or modify
 /// it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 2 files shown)


c-lipka:
I don't know if it's worth optimizing the actual TTF raytracing code; after all, it is redundant anyway, as we could instead convert any `text` objects to cubic spline `prism` primitives internally.

wfpokorny:
While long term I agree with 'core-prism' code long term for prisms, fonts and eventual svg (spline) support, I see less than complete redundancy of text objects with prisms - especially in the near term.  

For example, by far the biggest gains - the ones which can easily be 2x and more for what I'm doing - come from the range pre-filter on the text string as 'generated glyph ' during the point inside_glyph checking. The range pre-filter comes with a penalty at one glyph (1 character) and ever larger gains as the number of glyphs/characters increases. The penalty is generally not attractive for the prisms I've used or those imported from say inkscape. Might we do something like create internal single character text objects within text objects (adding these the the search structures) on the fly - I suppose, but I'm not up for that work any time soon. Doing some of the pre-filter work on text object creation at the expense of storage might work too, but gets complicated and I want to do other work.  

Is the sweep method compatible - I wasn't sure. A text object built on the prism code should perhaps enable the prism taper / clamp in z capability.  

I look at the truetype code and I think we should have on the list to gut the entirety of it and use instead  some other open source package for the front end - freetype.org or that popular package on github(1) the name of which escapes me? As you know a lot is not properly supported in the code for modern fonts / additional font formats. 

(1) - This package has some 2D rectangle packing to font characters code which could make for very fast inside_object tests and which I think would work well with our current string to glyph method. I've lots of ideas :-), not enough time and sadly C++ skills which won't be up to bigger refactoring jobs for a long while yet.          

Sorry for getting long-winded. Please feel free to use parts (normal fix is simple?) or to close this pull request entirely. I'll be using this code no matter myself pending anything better for other work I want to do. 

Aside: Perhaps we should open some general area of discussion issues as a place to discuss long term ideas? "Text object font support," "svg support," "replacement editor," and so on. Might let us close out smaller sub issues for subject matter specific ones maintaining a list of ideas, feature requests all in one place.

wfpokorny:
Closing this request as normal fix manually picked up by Christoph and mostly to avoid pointless build checks as I rebase for my own work. Code clean up and performance portion can be addressed - or not - after 3.8 released and master moves to 3.8.1.

Issue #285, opened 2017-04-30
Port of FS243 - Sphere sweep behaves wrong when scaled


wfpokorny:
http://bugs.povray.org/task/243  (Simon (infoised))

(When scaled with identical negative values for x,y,z as in scale -1)

(also sphere_sweep related: #59, #147 #287)

---------------
Details:

The sphere_sweep renders well when specified directly, but when it is scaled, its bounding box is calculated incorrectly, which clips the object so it almost disappears.

The effect is present for all three types of splines.

I'm attaching a test scene and the rendering result. The saving of the object with #declare has no effect, I just wanted to show both transformed and untransformed version.

I don't think this issue is related to other artifacts occurring with sphere_sweep, as it is obviously an issue of the internal bounding box.
   fig_bug_sweep.png (7 KiB)
   fig_bug_sweep.pov (0.7 KiB)

See zip file:
[FS243.zip](https://github.com/POV-Ray/povray/files/966768/FS243.zip)

---------------
Comments:

---
Comment by Thorsten Fröhlich (thorsten) - Monday, 28 May 2012, 00:03 GMT+5

The bounding of spline-based sphere sweeps never worked right and it known to have many bugs, in particular because the bounding box does not correctly consider splines were the curve extends beyond the control points. Transformations just make the problems more apparent. The best solution is to manually bound sphere-sweeps.

---
Comment by Simon (infoised) - Monday, 25 February 2013, 21:27 GMT+5

Found the error. It's an easy fix. The SphereSweep::Scale has a special case for scaling if scaling is uniform: it simply multiplies all vectors AND ALL RADII by the scaling factor. This makes radius negative for negative scaling factors, which makes the bounding box shrink instead of grow. The code should be

```
void SphereSweep::Scale(const VECTOR Vector, const TRANSFORM *tr)
{
if((Vector[X] != Vector[Y]) || (Vector[X] != Vector[Z]))
{
        if(Trans == NULL)
                Trans = Create_Transform();
}

if(Trans == NULL)
{
        for(int i = 0; i < Num_Modeling_Spheres; i++)
        {
                VScaleEq(Modeling_Sphere[i].Center, Vector[X]);
                Modeling_Sphere[i].Radius *= fabs(Vector[X]); /*missing absolute value*/
        }
        Compute();
        Compute_BBox();
}
else
Transform(tr);
}
```

This fixes the bounding box for linear_spline (linear splines never stick out of their bounding box, so they don't have the same problems as the other types).

[...]

PR #286, opened 2017-04-30
Adopting Simon's fix for sphere_sweep scaling issue #285.

wfpokorny:
Reported originally as flyspray issue FS243.


c-lipka:
The proposed fix breaks down for cases where the user deliberately specifies some radii as positive and some as negative for effect (creating a "pinch" in between the corresponding control points), as it will force all radii to have the same sign in such a case.

I've checked in a different solution by now.

PR #288, opened 2017-05-02
Restoring 3.6 user bounding control.

wfpokorny:
The ini options Bounding=off, Bounding_Threshold=n and
the command line options -MB, -MBn turning automatic
bounding off will again function.

This commit fixes the remaining bounding control issues reported in issue #59.

The following test scene when run with -mb will now render the badly auto-bounded cubic sphere_sweep correctly.

```
#version 3.7;

// povray -mb <this_pov_scene_filename>
//
// Note: This test depends on the sphere_sweep generating
// bad bounding so it only renders correctly if POV-Ray
// bounding is completely off.
//

global_settings { assumed_gamma 1 }
camera {
    location <0.0, 1.5, -8.5>
    look_at <0.0, 1.5, 0.0>
    angle 29
}
light_source { <-4.562, 10.625, -7.902>, rgb 1 }

#declare R = 0.07;

sphere_sweep {
    cubic_spline 6,
    <1, -1>, R,
    <-1, -1>, R,
    <-1, 1>, R,
    <1, 1>, R,
    <1, -1>, R,
    <-1, -1>, R
    pigment { red 1 }
    translate <0,1.5,0>
}

sphere { <0.0, 1.9, 0.0>, 0.05 pigment { green 1 } }
sphere { <0.0, 1.7, 0.0>, 0.05 pigment { green 1 } }
sphere { <0.0, 1.5, 0.0>, 0.05 pigment { green 1 } }
sphere { <0.0, 1.3, 0.0>, 0.05 pigment { green 1 } }
sphere { <0.0, 1.1, 0.0>, 0.05 pigment { green 1 } }
```

### DIFF
--- a/source/core/render/trace.cpp
+++ b/source/core/render/trace.cpp
@@ -8,7 +8,7 @@
 /// @parblock
 ///
 /// Persistence of Vision Ray Tracer ('POV-Ray') version 3.7.
-/// Copyright 1991-2016 Persistence of Vision Raytracer Pty. Ltd.
+/// Copyright 1991-2017 Persistence of Vision Raytracer Pty. Ltd.
 ///
 /// POV-Ray is free software: you can redistribute it and/or modify
 /// it under the terms of the GNU Affero General Public License as
@@ -339,12 +339,13 @@ bool Trace::FindIntersection(ObjectPtr object, Intersection& isect, const Ray& r
         BBoxVector3d invdir;
         BBoxDirection variant;
 
-        Vector3d tmp(1.0 / ray.GetDirection()[X], 1.0 / ray.GetDirection()[Y], 1.0 /ray.GetDirection()[Z]);
+        Vector3d tmp(1.0 / ray.GetDirection()[X], 1.0 / ray.GetDirection()[Y], 1.0 / ray.GetDirection()[Z]);
         origin = BBoxVector3d(ray.Origin);
         invdir = BBoxVector3d(tmp);
         variant = (BBoxDirection)((int(invdir[X] < 0.0) << 2) | (int(invdir[Y] < 0.0) << 1) | int(invdir[Z] < 0.0));
 
-        if(object->Intersect_BBox(variant, origin, invdir, closest) == false)
+        // Note: At leaf when called. Suspect not strictly needed. However, without slightly slower in performance.
+        if((object->Intersect_BBox(variant, origin, invdir, closest) == false) && (sceneData->boundingMethod != 0))
             return false;
 
         if(object->Bound.empty() == false)
@@ -392,12 +393,13 @@ bool Trace::FindIntersection(ObjectPtr object, Intersection& isect, const Ray& r
         BBoxVector3d invdir;
         BBoxDirection variant;
 
-        Vector3d tmp(1.0 / ray.GetDirection()[X], 1.0 / ray.GetDirection()[Y], 1.0 /ray.GetDirection()[Z]);
+        Vector3d tmp(1.0 / ray.GetDirection()[X], 1.0 / ray.GetDirection()[Y], 1.0 / ray.GetDirection()[Z]);
         origin = BBoxVector3d(ray.Origin);
         invdir = BBoxVector3d(tmp);
         variant = (BBoxDirection)((int(invdir[X] < 0.0) << 2) | (int(invdir[Y] < 0.0) << 1) | int(invdir[Z] < 0.0));
 
-        if(object->Intersect_BBox(variant, origin, invdir, closest) == false)
+        // Note: At leaf when called. Suspect not strictly needed. However, without slightly slower in performance.
+        if((object->Intersect_BBox(variant, origin, invdir, closest) == false) && (sceneData->boundingMethod != 0))
             return false;
 
         if(object->Bound.empty() == false)
--- a/source/core/render/trace.cpp
+++ b/source/core/render/trace.cpp
@@ -331,6 +331,20 @@ bool Trace::FindIntersection(Intersection& bestisect, const Ray& ray, const RayO
     return false;
 }
 
+
+// The following Trace::FindIntersection functions are never called on
+// bounding hierarchy leaves; rather only called when not traversing the
+// bounding hierarchy:  Either to compute the intersections with infinite
+// objects if they're not already accounted for by the bounding hierarchy
+// chosen, or because bounding is disabled entirely, or because only the
+// intersections with one particular object are of interest (e.g.  the
+// projected_through object of a light source, but also the shadow caching
+// mechanism).
+//
+// In the former two cases it wouldn't be necessary to perform the bounding
+// box test, but in the latter case it is indispensable for good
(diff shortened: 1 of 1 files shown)


wfpokorny:
Christoph, I looked at the travis-cl fail here, but it looks to me like it might have been something internal to a travis-ci script.

c-lipka:
I'm worried that this change might fall short: Aside from `FindIntersection()` in `trace.cpp`, there are other places in the code that also use `Intersect_BBox()` without checking the `boundingMethod` parameter, most notably the `Find_Intersection()` set of functions in `object.cpp` (which do not have access to the scene data).

source/core/render/trace.cpp
```diff
@@ -339,12 +339,13 @@ bool Trace::FindIntersection(ObjectPtr object, Intersection& isect, const Ray& r
...
         BBoxDirection variant;
 
-        Vector3d tmp(1.0 / ray.GetDirection()[X], 1.0 / ray.GetDirection()[Y], 1.0 /ray.GetDirection()[Z]);
+        Vector3d tmp(1.0 / ray.GetDirection()[X], 1.0 / ray.GetDirection()[Y], 1.0 / ray.GetDirection()[Z]);
         origin = BBoxVector3d(ray.Origin);
         invdir = BBoxVector3d(tmp);
         variant = (BBoxDirection)((int(invdir[X] < 0.0) << 2) | (int(invdir[Y] < 0.0) << 1) | int(invdir[Z] < 0.0));
 
-        if(object->Intersect_BBox(variant, origin, invdir, closest) == false)
+        // Note: At leaf when called. Suspect not strictly needed. However, without slightly slower in performance.
```
c-lipka: As a matter of fact, this function is _never_ called on bounding hierarchy leaves; it is only called when _not_ traversing the bounding hierarchy: Either to compute the intersections with infinite objects if they're not already accounted for by the bounding hierarchy chosen, or because bounding is disabled entirely, or because only the intersections with one particular object are of interest (e.g. the `projected_through` object of a light source, but also the shadow caching mechanism).

In the former two cases it wouldn't be necessary to perform the bounding box test, but in the latter case it is indispensable for good performance.

source/core/render/trace.cpp
```diff
@@ -392,12 +393,13 @@ bool Trace::FindIntersection(ObjectPtr object, Intersection& isect, const Ray& r
...
```
c-lipka: See above.

source/core/render/trace.cpp
```diff
@@ -339,12 +339,13 @@ bool Trace::FindIntersection(ObjectPtr object, Intersection& isect, const Ray& r
...
+        if((object->Intersect_BBox(variant, origin, invdir, closest) == false) && (sceneData->boundingMethod != 0))
```
c-lipka: Due to the bounding method test being significantly less expensive than the bounding box intersection test, I would lean towards swapping around the two tests.

source/core/render/trace.cpp
wfpokorny: Thanks for digging and explaining in detail. I didn't understand why this bbox test helped performance but the use in the latter two situations explains it. I'll update the 'Note:' in one of the two functions to be more or less your comment & given it's longish, reference the note in the other.

source/core/render/trace.cpp
wfpokorny: Hmm. Functionally the same so I have no objection. However, I intentionally ordered it with the bounding method test second because the test in normal use(1) will be true and unimportant when the bbox test is false. In other words, in normal use both the bounding method test and the bbox test will have to be done, if the method test is first, while ordered as it is, a lazy evaluation is possible when the bbox test is false. 

Do you still want to change the order? 

(1) I see the -mb mode as useful while debugging issues, but very useful otherwise.

c-lipka:
With the sphere sweep bounding sorted out by now, I think the `-MB` issue has become less pressing, so we don't necessarily need to hurry it into 3.7.1.

Do you agree? If so, I'd like to drop this pull request from the 3.7.1 milestone for now. (If we manage to finalize this pull request soon enough, we can still put it back in again.)

wfpokorny:
Yes, I'm OK dropping this as a 3.7.1 milestone. 

The -mb mode is useful for more than bad initial bounding. One of my pet projects over the past month has been cleaning up user bounding +ur / -ur with an eye toward performance - perhaps helping too isosurfaces.  Today our default is supposedly +ur, but in fact where the user bounding is ignored in the frame set up code we exempt a small set of objects for reasons lost on me. There are too some secondary issues with some odd - save the user from themselves - volume testing I think we should lose. In other words there is no way today to remove user bounding in some situations except with a working -mb.

There is too an untracked issue I believe with how sphere_sweep user bounding is adjusted by downstream transforms. Some Sean Day found it and you dug into it a little in 2013. I've hacked together some code added to  SphereSweep::Compute_BBox just after the existing bounding code that fixes it, but I've not pushed on this issue because I don't understand yet whether the downstream transform issue for user bounding is unique to the sphere_sweep or something more general..? Also think something like my code is cleaner than what we have in parser.cpp today and perhaps it should be some common utility function called in both places. 

The code added to SphereSweep::Compute_BBox as I have it today for the record: 

```
     if(!this->Bound.empty() || !this->Clip.empty())
    {
        Vector3d tmpMins = Vector3d(BOUND_HUGE);
        Vector3d tmpMaxs = Vector3d(-BOUND_HUGE);
                 mins    = Vector3d(BOUND_HUGE);  // To use user bounding as specified.
                 maxs    = Vector3d(-BOUND_HUGE);

        for(vector<ObjectPtr>::iterator Sib = this->Bound.begin(); Sib != this->Bound.end(); Sib++)
            if(!Test_Flag((*Sib), INVERTED_FLAG))
                {
                    Make_min_max_from_BBox(tmpMins, tmpMaxs, (*Sib)->BBox);
                    mins = min(mins, tmpMins);
                    maxs = max(maxs, tmpMaxs);
                }

        if(this->Clip != this->Bound)
            for(vector<ObjectPtr>::iterator Sib = this->Clip.begin(); Sib != this->Clip.end(); Sib++)
                if(!Test_Flag((*Sib), INVERTED_FLAG))
                    {
                        Make_min_max_from_BBox(tmpMins, tmpMaxs, (*Sib)->BBox);
                        mins = min(mins, tmpMins);
                        maxs = max(maxs, tmpMaxs);
                    }

        mins = max(mins, Vector3d(-BOUND_HUGE/2)); // Clamp to BOUND_HUGE range.
        maxs = min(maxs, Vector3d(BOUND_HUGE/2));
    }

    Make_BBox_from_min_max(BBox, mins, maxs);
```

wfpokorny:
With respect to my previous comments about the untracked issue with how sphere_sweep user bounding is adjusted on a transform, please see issue #295 and ignore my rambling above.

wfpokorny:
Closing per request to drop as a 3.7.1 milestone. I'll open a new master based request with the comment and code changes you asked for from this request later today.

PR #296, opened 2017-05-22
Make build reproducible

bmwiedemann:
when linking binaries with g++, order of input files
influences ordering of functions in the output

See https://reproducible-builds.org/ for why this matters.

### DIFF
--- a/unix/prebuild.sh
+++ b/unix/prebuild.sh
@@ -778,7 +778,7 @@ case "$1" in
   ;;
 
   *)
-  files=`find $dir -name "*.cpp" -or -name "*.h" | sed s,"$dir/",,g`
+  files=`find $dir -name "*.cpp" -or -name "*.h" | sed s,"$dir/",,g | sort`
 
   echo "Create $makefile.am"
   cat Makefile.header > $makefile.am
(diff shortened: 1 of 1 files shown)


c-lipka:
I don't really see how reproducible builds might be relevant for POV-Ray.

Also, see https://github.com/bitcoin/bitcoin/blob/master/doc/gitian-building.md for an example of what it _really_ takes to get reproducible builds.

bmwiedemann:
Yes, they were one of the first doing this.
For povray in openSUSE, it just takes this simple 2-line patch, which I already wrote and tested and has very low risk to break anything.
It is this simple, because we have meanwhile already patched many tools like gcc and rpm that would otherwise have inserted non-deterministic data.

This is why today 9670 out of 11715 packages in openSUSE are able to produce bit-identical binary rpms. And the goal is to have 100% - that obviously includes povray and we try to get patches upstream that make sense for non-openSUSE systems.

bmwiedemann:
So if you wanted to produce reproducible binaries for any of the 9670 good packages, you could do atm
```
osc checkout openSUSE:Factory python-Flask && cd $_
osc build --vm-type kvm --alternative-project home:bmwiedemann:reproducible
```

You can do that on different hosts at different times and get bit-identical results. When we progress more, it will get easier to also re-build the official builds.
But for now it is already an advantage when we get a new glibc that we do not always have to re-publish a new povray RPM and waste bandwidth of our mirrors and users.

Thus, please merge ;-)

c-lipka:
I doubt that building POV-Ray on _different_ hosts will get bit-identical results (even with your changes), as by default POV-Ray builds with the `-march=native` option.

But I guess I do get your point that when building POV-Ray for distribution via a package manager, you probably want to get the same result every time _at least_ if you have a well-defined build environment.

Issue #295, opened 2017-05-21
Inconsistencies in user bounding and  user bounding control. 


wfpokorny:
The following example, originally from Sean Day, works as expect using -ur but not with the default +ur though the specified bounded_by box meets the "better bounding" tests in parse.cpp. At the core the +ur mode is not making downstream transform adjustments that consider the bbox leading to confusing and inconsistent behavior. 

```
//
// -ur and the translated bounded_by box works as expected.
// With the +ur default the bounded_by box is used if not transformed  
// or it is transformed in the base object. 

#version 3.7;
global_settings { assumed_gamma 1 }
camera
{ location <0.0, 1.5, -8.5>
  look_at <0.0, 1.5, 0.0>
  angle 30
}
light_source { <-4.5, 10.5, -8.0>, rgb 1 }
box { <-10, 0, -10>, <10, 12, 10> hollow pigment { rgb 1 } }

#declare R = 0.07;
#declare Rp = R + 0.001;

union {
    sphere_sweep {
        cubic_spline 6,
        <1, -1>, R,
        <-1, -1>, R,
        <-1, 1>, R,
        <1, 1>, R,
        <1, -1>, R,
        <-1, -1>, R
        bounded_by { box { -<1.5, 0.5, Rp>, <1.5, 0.5, Rp> } }

        texture { pigment { red 1 } }
        interior_texture { pigment { color rgb <1,1,0> } }

//      translate 1.5 * y    // If translate done here things work. bbox set up in parse.cpp.
    }
    sphere { <-0.3, 0.0, 0.0>, 0.05 pigment { blue 1 } }
    translate 1.5 * y        // This should adjust the bounded_by "bbox" too when the original is
                             // not stripped by a volume test in parse.cpp...
                             // The bounded_by box is adjusted by Translate_Object in object.cpp
                             // But unless -ur is used or extra code to look at the bbox added to
                             // SphereSweep::Compute_BBox(), the adjustment is ignored.
}
sphere { <0.0, 1.9, 0.0>, 0.05 pigment { green 1 } }
sphere { <0.0, 1.7, 0.0>, 0.05 pigment { green 1 } }
sphere { <0.0, 1.5, 0.0>, 0.05 pigment { green 1 } }
sphere { <0.0, 1.3, 0.0>, 0.05 pigment { green 1 } }
sphere { <0.0, 1.1, 0.0>, 0.05 pigment { green 1 } }
```

[...]

PR #298, opened 2017-06-11
Restoring 3.6 user bounding control.

wfpokorny:
The ini options Bounding=off, Bounding_Threshold=n and the command line
options -MB, -MBn turning automatic bounding off will again function.

Originally a pull against release/3.7.1 as issue #288, but made a
request against master per Christoph.  Related to issue #295 and
necessary to fix and close the last of issue #59.


wfpokorny:
Branch at closing still exists and is occasionally re-based to master.

PR #299, opened 2017-06-12
Fix crash in radiosity.cpp related to mosaic preview.

c-lipka:
This pull request duplicates an old fix already applied to the 3.7.1 branch, fixing a crash in `radiosity.cpp` when mosaic pretrace is enabled (originally tracked on FlySpray as FS313 and recently reported again on the `povray.bugreports` newsgroup).


c-lipka:
Probably not relevant in v3.7.0 binaries, because those asserts should be disabled in release builds, but it may be wiser to fix it nonetheless.

PR #302, opened 2017-06-25
Cleanup indentation 

Mausy5043:
Cleaned up the indentation to improve readability.


c-lipka:
Various of the Autoconf macros we're using originate from (and, as I presume, are verbatim copies from) 3rd party sources, so I wouldn't want to tamper with those.

Also, we're no longer making any changes to POV-Ray 3.7.0 (which is what the `3.7-stable` branch currently reflects) except to fix severe bugs or compatibility issues with modern build environments. As a rule of thumb, any pull requests should be based on the `master` branch.

Mausy5043:
OK. No problem.

PR #305, opened 2017-07-06
Default ambient for 3.8 and above

LeForgeron:
When version is set as the first statement of the scene or via option, and the version is 3.8 or more, the default ambient is 0 instead of the previous 0.1 which has problem with the gamma handling.


wfpokorny:
:-) Typo - not yours - fixed and everything!

c-lipka:
Looks like you also addressed the INI / command-line version override already, so nothing to complain from my side. Presuming nobody finds other issues we've missed, I'm ready to pull as soon as the general go-ahead is given.

PR #308, opened 2017-07-10
Feature unix terminal, try2

eblanca:


unix/configure.ac
```diff
@@ -728,6 +728,9 @@ if test x"$C99_COMPATIBLE_RADIOSITY" != x"" || test x"$ac_cv_sizeof_int" != x"4"
   fi
 fi
 
+AC_HEADER_TIOCGWINSZ
```
c-lipka: It seems a bit fishy to me that you're using the `AC_HEADER_TIOCGWINSZ` in the autoconf script, but not testing `GWINSZ_IN_SYS_IOCTL` in the C++ code. Please double-check.

c-lipka:
Glad to see the pull request get past the Travis CI build tests now.

eblanca:
With this commit the macro TIOCGWINSZ will be found even on those systems declaring it into termios.h

c-lipka:
I think the general idea is vindicated, but I still see some cleaning up that should be done.

For starters, `GetTerminalWidth()` is clearly a platform-specific function, and therefore should be moved from the platform-neutral code portion (`source/*`) to the platform-specific portion (traditionally `vfe/[unix|windows]/vfeplatform.cpp`, though `platform/[unix|windows]/syspov*` should be preferred now; `syspovconsole.*` would do nicely I think); this will also solve the problem of `<termios.h>` being unavailable on Windows in a more natural manner.

Also, currently the `GetTerminalWidth()` call completely overrides any value passed into the constructor parameter `wrapwidth` (even when `GetTerminalWidth()` can't come up with an idea of its own). Actually I think the `TextStreamBuffer` class is too low-level to take this matter into its own hands (it might be used for purposes entirely unrelated to console output, after all), and that `GetTerminalWidth()` should be invoked by the constructor's caller instead (or maybe that caller's caller; I'm not sure how deep that rabbit-hole actually is).

That said, thanks for your persistence so far; even if you'd prefer to leave the code as it is, I guess I'd be inclined to pull it and build upon it. But if you'd like to continue honing it and don't mind me finding more nits to pick, that'd be even be more appreciated.

eblanca:
Thanks Christoph, I must tell the povray source base is so huge I find quite hard to move across.
Having said that, yes, I find reasonable to move `GetTerminalWidth()` into a more proper module.

> currently the `GetTerminalWidth()` call completely overrides

Yes, it was done only for the purpose of getting the function call working well, assuming a further work to be done once this was reached.

Indeed, I will put some effort in getting clearer code.

> thanks for your persistence

No pun intended with our preferred raytracer, eh?

c-lipka:
> No pun intended with our preferred raytracer, eh?

Actually, no, indeed not. But I'm proud of my subconscious. ;)

Issue #317, opened 2017-09-02
Boost 1.65.0 incompatibility


ilovezfs:
As of Boost 1.65.0, the tr1 headers have all been removed. However, povray master is still unconditionally including them, as here: https://github.com/POV-Ray/povray/blob/master/unix/povconfig/syspovconfig.h#L59

This renders povray incompatible with Boost >= 1.65.0. It would be great if you could add compatibility for the current version of Boost. Thanks!

fxcoudert:
Patch fixing the issue: [tr1.diff.txt](https://github.com/POV-Ray/povray/files/1271988/tr1.diff.txt)

c-lipka:
Please test the `hotfix/issue_317` branch, which should fix the issue and is scheduled to be merged back into the `release/v3.8.0` branch (which in turn will be merged back into the `master` branch).

ilovezfs:
@c-lipka looks like 3.7.0.4 is good. There's a minor header issue. I think it should `#include <cstdlib>` in `source/frontend/processrenderoptions.cpp`. The errors are

```
/usr/bin/clang++ -std=c++11 -stdlib=libc++ -DHAVE_CONFIG_H -I. -I..  -I.. -I../source/backend -I../source/base -I../source/frontend -I../unix -I../vfe -I../vfe/unix -F/usr/local/Frameworks -pthread -I/usr/local/opt/boost/include  -pipe -Wno-multichar -Wno-write-strings -O3 -ffast-math -march=native -Os -w -pipe -march=native -mmacosx-version-min=10.11 -D_THREAD_SAFE -c -o frontend/processrenderoptions.o frontend/processrenderoptions.cpp
frontend/processrenderoptions.cpp:463:59: error: use of undeclared identifier 'atof'
                                                err = POVMSUtil_SetFloat(&decobj, kPOVAttrib_Value, atof(ptr));
                                                                                                    ^
frontend/processrenderoptions.cpp:516:15: error: use of undeclared identifier 'atof'
                        floatval = atof(param);
                                   ^
2 errors generated.
make[2]: *** [frontend/processrenderoptions.o] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
```

ilovezfs:
@c-lipka full build log is here: https://gist.github.com/ilovezfs/d4f1368f5473fd1ae26a53d41e1b751a

c-lipka:
[...]

PR #318, opened 2017-09-02
Remove boost TR1

fxcoudert:
Fixes incompatibility with recent boost versions https://github.com/POV-Ray/povray/issues/317
cc @ilovezfs


wfpokorny:
First, It happens many of the core developers are occupied with other than POV-Ray at the moment so response might be slow. 

I was not involved in the original implementation, but I know from other discussions the 'belief' was the ' #include <boost/tr1/*' style would just work when the code implementation showed up in the base compiler while not breaking those with older installs and compilers.  

However, digging a little just now I found: 

http://www.boost.org/doc/libs/1_64_0/doc/html/boost_tr1/usage.html#boost_tr1.usage.native

which looks to me like we should make the changes you've made but also update the *nix build system for another boost include in:  boost-install-path/boost/tr1/tr1 as a start for compatibility. Looks like that tr1/tr1 include exists back to at least boost 1.58. 

I know our boost m4 macros - which are just copied into <source_install>povray/unix/config with an ax_prefix are somewhat out of date with current ones, but I also see nothing in my copies from March which would add such an include by default. So...   

While it is a Holiday weekend here, I've got other commitments. Would you be willing to hack your  branch <source_install>povray/unix/configure.ac file in the boost section so the line: 

  `CPPFLAGS="$CPPFLAGS $BOOST_CPPFLAGS"  # append`

reads: 

  `CPPFLAGS="$CPPFLAGS $BOOST_CPPFLAGS -I/usr/include/boost/tr1/tr1"  # append`

? 

After which, test if your compiles still work and if so, commit and push to this pull request to see if it fixes any of the backward compatibility issues in the github integration tests.  

OK if not. Just might be a while until someone else can look at this. 

Suppose even if adding this include works, it's not what we should do long term given, for one, the boost includes might not always be in that specific location.  The link above also lists other issues.  Must be too the tr1 versions just don't exist in 1.65 onward which implies boost forcing the use of compilers with inbuilt support I guess - at least with base boost installs? Wonder if the tr1 stuff can still be installed specifically?

c-lipka:
It appears that boost/tr1 has been deprecated since boost 1.56, which was released in August 7, 2014.

In my optinion the removal of that library was a poor choice, as it removes a crucial piece of the puzzle to keep smart-pointer-using code compatible across C++ standard generations.

For our current codebase (previously referred to as 3.7.1, but redesignated 3.8.0 by now), I think this is the point where we should switch to C++11 entirely.

For our legacy codebase (3.7.0.x) however, I'm not sure if that would be the right choice. Aiming for compatibility with C++11 (and C++14 and whatnot) and modern boost versions is one thing; ditching compatibility with compilers that were up-to-date when 3.7.0 was originally developed is an entirely different thing.

In my opinion, the proposed changes are therefore not an option. We should go for something smarter.

ilovezfs:
>For our current codebase (previously referred to as 3.7.1, but redesignated 3.8.0 by now), I think this is the point where we should switch to C++11 entirely.

This PR is against master, which I assume is the 3.8.0 branch, no?

c-lipka:
Sorry, skimming over this pull request I got mislead by the homebrew reference citing 3.7.0.3.

`master` is actually not the 3.8.0 branch; instead, it is supppsed to "inherit" from `release/v3.8.0`.

I had already prepared a set of changes to the modern (post-3.7.0) codebase to make the transition to C++11. I'll have a look whether I already included the removal of boost/tr1 dependency.

ilovezfs:
@c-lipka Ah, OK. For 3.7.0.3 in Homebrew, we needed a modified version of this PR, which is here: https://github.com/Homebrew/formula-patches/pull/173.

c-lipka:
See commit 1dbca0b.

PR #323, opened 2017-10-09
GetTerminalWidth() into a more proper module

eblanca:
Following your suggestions, I think now the original idea is pretty realized.
Only a question remains for me, and it is about that `vfesession.cpp`, does it only get built under unix-like OSs, does it?


eblanca:
I had my answer, it gets.
I hope I fixed it.

PR #336, opened 2017-11-07
Feature extended ovus

LeForgeron:
Just add/update the small change of 21 Oct 2017, to have the same orientation on ovus than on lemon for uv-mapping.


LeForgeron:
Forget that pull request, doing it again from fresh branch from release/v3.8.0

PR #338, opened 2017-11-19
fix the orientation of new uv-mapping of ovus to be similar to cylind…

LeForgeron:
…er & lemon

(it was upside down)


LeForgeron:
Redone in #355

PR #343, opened 2018-01-14
Clean up use of `snprintf`/`vsnprintf` to use C++11 standard functions.

c-lipka:
Also replace `FILE_NAME_LENGTH` macro along the way in hope to clarify the code, and eliminate `MAX_PATH` from all but the Windows-specific code.

### DIFF
--- a/source/base/configbase.h
+++ b/source/base/configbase.h
@@ -10,7 +10,7 @@
 /// @parblock
 ///
 /// Persistence of Vision Ray Tracer ('POV-Ray') version 3.8.
-/// Copyright 1991-2017 Persistence of Vision Raytracer Pty. Ltd.
+/// Copyright 1991-2018 Persistence of Vision Raytracer Pty. Ltd.
 ///
 /// POV-Ray is free software: you can redistribute it and/or modify
 /// it under the terms of the GNU Affero General Public License as
@@ -678,6 +678,17 @@
     // The following two lines work around doxygen being unable to document undefined macros.
     #define POV_SYS_IMAGE_EXTENSION (undefined)
     #undef POV_SYS_IMAGE_EXTENSION
+#endif
+
+/// @def POV_FILENAME_BUFFER_CHARS
+/// The number of characters to reserve for file name buffers.
+///
+/// This setting is used in allocating temporary buffers to construct file names, and should be set
+/// to the maximum number of ASCII characters in a file name the system can handle safely.
+/// The value is understood to exclude any terminating NUL character.
+///
+#ifndef POV_FILENAME_BUFFER_CHARS
+    #define POV_FILENAME_BUFFER_CHARS 199
 #endif
 
 /// @def POV_PATH_SEPARATOR
--- a/source/base/fileinputoutput.cpp
+++ b/source/base/fileinputoutput.cpp
@@ -8,7 +8,7 @@
 /// @parblock
 ///
 /// Persistence of Vision Ray Tracer ('POV-Ray') version 3.8.
-/// Copyright 1991-2017 Persistence of Vision Raytracer Pty. Ltd.
+/// Copyright 1991-2018 Persistence of Vision Raytracer Pty. Ltd.
 ///
 /// POV-Ray is free software: you can redistribute it and/or modify
 /// it under the terms of the GNU Affero General Public License as
@@ -38,6 +38,7 @@
 
 // C++ variants of standard C header files
 #include <cstdarg>
+#include <cstdio>
 #include <cstdlib>
 
 // Standard C++ header files
@@ -301,7 +302,7 @@ void OStream::printf(const char *format, ...)
     char buffer[1024];
 
     va_start(marker, format);
-    vsnprintf(buffer, 1023, format, marker);
+    std::vsnprintf(buffer, sizeof(buffer), format, marker);
     va_end(marker);
 
     write(reinterpret_cast<const void *>(buffer), strlen(buffer));
--- a/source/parser/parser.cpp
+++ b/source/parser/parser.cpp
@@ -8,7 +8,7 @@
 /// @parblock
 ///
 /// Persistence of Vision Ray Tracer ('POV-Ray') version 3.8.
-/// Copyright 1991-2017 Persistence of Vision Raytracer Pty. Ltd.
+/// Copyright 1991-2018 Persistence of Vision Raytracer Pty. Ltd.
 ///
 /// POV-Ray is free software: you can redistribute it and/or modify
 /// it under the terms of the GNU Affero General Public License as
@@ -39,6 +39,8 @@
 // C++ variants of C standard header files
 #include <cctype>
 #include <cmath>
+#include <cstdarg>
+#include <cstdio>
 #include <cstdlib>
 
 // C++ standard header files
@@ -179,7 +181,7 @@ void Parser::Run()
 {
     int         error_line = -1;
     int         error_col = -1;
-    UCS2String  error_filename(MAX_PATH, 0); // Pre-claim some memory, so we can handle an out-of-memory error.
+    UCS2String  error_filename(POV_FILENAME_BUFFER_CHARS, 0); // Pre-claim some memory, so we can handle an out-of-memory error.
     POV_LONG    error_pos = -1;
 
     try
@@ -253,7 +255,7 @@ void Parser::Run()
             if (Token.FileHandle != NULL)
             {
                 // take a (local) copy of error location prior to freeing token data
-                // NB error_filename has been pre-allocated for strings up to _MAX_PATH
+                // NB error_filename has been pre-allocated for strings up to POV_FILENAME_BUFFER_CHARS
                 error_filename = Token.FileHandle->name();
                 error_line = Token.Token_File_Pos.lineno;
                 error_col = Token.Token_Col_No;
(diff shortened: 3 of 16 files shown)


source/base/configbase.h
```diff
@@ -680,6 +680,17 @@
...
+/// The number of characters to reserve for file name buffers.
+///
+/// This setting is used in allocating temporary buffers to construct file names, and should be set
+/// to the maximum number of ASCII characters in a file name the system can handle safely.
+/// The value is understood to exclude any terminating NUL character.
+///
+#ifndef POV_FILENAME_BUFFER_CHARS
+    #define POV_FILENAME_BUFFER_CHARS 199
+#endif
+
```
c-lipka: The newly introduced system-specific compile-time configuration setting `POV_FILENAME_BUFFER_CHARS` replaces `FILE_NAME_LENGTH`, to achieve the following goals:
- Clarify whether the size does or does not include room for a terminating NUL character. (The old macro de-facto did, but this was not documented anywhere, and the name might have been interpreted otherwise; the new macro is clearly documented, and the name chosen accordingly, to exclude any terminating NUL character. Net size was favoured over gross size to better fit with C++-style strings.)
- Clarify whether the size should reflect the system's _absolute_ limit for file name size, the _safe_ limit, or something else entirely. (The new macro is now specified to be just the size for internal buffers in POV-Ray; using the maximum safe limit is just a recommendation.)
- Clarify how the macro's value should be interpreted in wide or multi-byte character contexts. (With the new macro no longer strictly tied to system file name limits, this should now be less of an issue.)

source/base/fileinputoutput.cpp
```diff
@@ -38,6 +38,7 @@
 
 // C++ variants of standard C header files
 #include <cstdarg>
+#include <cstdio>
```
c-lipka: `<cstdio>` is explicitly included here (and in other places) as `std::vsnprintf` is declared there.

source/base/fileinputoutput.cpp
```diff
@@ -301,7 +302,7 @@ void OStream::printf(const char *format, ...)
     char buffer[1024];
 
     va_start(marker, format);
-    vsnprintf(buffer, 1023, format, marker);
+    std::vsnprintf(buffer, sizeof(buffer), format, marker);
```
c-lipka: Previously, (almost) all invocations of `vsnprintf` and `snprintf` passed the buffer capacity _minus 1_, i.e. _excluding_ room for the terminating NUL character, despite the standard C99 functions expecting the buffer capacity to _include_ room for the NUL. This was done because we were de-facto using `_vsnprintf` and `_snprintf` on Windows machines, which did not follow the standard, and this pattern of invocation was considered safe either way (although on Windows machines it relied on the buffer being zero-initialized).
As we're now using the C++11 functions `std::vsnprintf` and `std::snprintf`, which follow the C99 convention even on Windows machines, we can now safely pass the full size of the buffer.

Also, instead of a literal number we're now using `sizeof(buffer)` wherever `buffer` is a fixed-size char array.

source/parser/parser.cpp
```diff
@@ -39,6 +39,8 @@
 // C++ variants of C standard header files
 #include <cctype>
 #include <cmath>
+#include <cstdarg>
```
c-lipka: `<cstdarg>` is explicitly included here (and in other places) as `va_start` and `va_end` are declared there.

source/parser/parser.cpp
```diff
@@ -179,7 +181,7 @@ void Parser::Run()
 {
     int         error_line = -1;
     int         error_col = -1;
-    UCS2String  error_filename(MAX_PATH, 0); // Pre-claim some memory, so we can handle an out-of-memory error.
+    UCS2String  error_filename(POV_FILENAME_BUFFER_CHARS, 0); // Pre-claim some memory, so we can handle an out-of-memory error.
```
c-lipka: Technically this change reduces the size of the buffer by 1. However, the `UCS2String` constructor expects the parameter to exclude the terminating NUL character, so the new code should more accurately reflect the original intention.

source/parser/parser_strings.cpp
```diff
@@ -226,13 +227,13 @@ UCS2 *Parser::Parse_Str(bool pathname)
     // TODO: consider changing to %g rather than %f for large numbers (e.g.
     // anything over 1e+64 for example). for now, we will only use %g if the
     // snprintf filled the buffer.
-    // NB test for < 0 is because stupid windows _snprintf can return negative values.
-    if (((l = snprintf(temp4, sizeof(temp4) - 1, temp3, val)) >= sizeof(temp4) - 1) || (l < 0))
+    // NB `snprintf` may report errors via a negative return value.
```
c-lipka: Windows' `_snprintf` wasn't stupid - the C99 standard (and by extension C++11) also allows `snprintf` to return a negative value, namely to indicate error conditions (presumably others than buffer overflows).

unix/configure.ac
```diff
@@ -613,23 +613,6 @@ AC_LANG_POP([C])
 #AC_FUNC_STRFTIME  # FIXME
 #AC_FUNC_VPRINTF   # FIXME
 
-# vsnprintf <stdarg.h> (C99)
-AC_CHECK_FUNC([vsnprintf],
```
c-lipka: As we now require C++11 anyway, availability of `std::vsnprintf` is guaranteed.

windows/povconfig/syspovconfig.h
```diff
@@ -227,10 +227,7 @@ namespace pov_base
 #define POV_NEW_LINE_STRING                 "\r\n"
 #define POV_SYS_IMAGE_EXTENSION             ".bmp"
 #define POV_SYS_IMAGE_TYPE                  BMP
-#define vsnprintf                           _vsnprintf
-#define snprintf                            _snprintf
-#define FILE_NAME_LENGTH                    _MAX_PATH
-#define POV_NAME_MAX                        _MAX_FNAME
```
c-lipka: `POV_NAME_MAX` was no longer used anywhere.

unix/povconfig/syspovconfig.h
```diff
@@ -102,21 +102,25 @@ using boost::intrusive_ptr;
...
-# define FILE_NAME_LENGTH   _POSIX_PATH_MAX
+    // Otherwise, use the most restrictive limit allowed by POSIX if defined.
+    // NB: _POSIX_PATH_MAX is understood including a terminating NUL character.
+    #define POV_FILENAME_BUFFER_CHARS   (_POSIX_PATH_MAX-1)
 #else
-# define FILE_NAME_LENGTH   200
+    // As a fallback, use an even more conservative limit.
+    #define POV_FILENAME_BUFFER_CHARS   199
 #endif
-#define MAX_PATH FILE_NAME_LENGTH  // FIXME: remove later
```
c-lipka: `MAX_PATH` has been eliminated from the platform-neutral code. It is still used in Windows-specific code, but only to set up buffers for API calls that are guaranteed to never write more than `MAX_PATH` bytes to the buffer.

vfe/vfe.cpp
```diff
@@ -547,7 +550,7 @@ void vfeProcessRenderOptions::ParseError(const char *format, ...)
   va_list marker;
 
   va_start(marker, format);
-  vsnprintf(str, sizeof(str)-2, format, marker);
```
c-lipka: Not sure why this is `sizeof(str)-2`, rather than `sizeof(str)-1`; I presume it to be a mistake. If anyone can think of any good reason to leave room for a trailing NUL character _and_ one more byte, let me know.

source/base/fileinputoutput.cpp
LeForgeron: Danger, vsnprintf does NOT put a terminating null when the result need to be truncated due to length constraint. (hence the 1023 used previously for buffer[1024]).

Both codes (before & after) are bogus, because it is missing a buffer[1023]='\0'; (before va_start, if vsnprintf() is called with sizeof(buffer)-1 ) to handle the situation

source/parser/parser_strings.cpp
LeForgeron: Is it a total removal of the line of code ? (missing more context, might be fine, just checking)

LeForgeron: fine, as you are testing the return value of snprintf to alter the answer in case of truncation.

Are we just sure that temp4 can hold "<invalid>" ? (always)

source/base/fileinputoutput.cpp
c-lipka: What you're writing was (and still is) true for Microsoft's implementation of `_vsnprintf` (note the leading underscore). However, with the introduction of C++11 Microsoft has also introduced a `vsnprintf` function, which (at least according to their docs) _does_ always NUL-terminate the string (as mandated by the C99 standard, which C++11 in turn refers to).

source/parser/parser_strings.cpp
c-lipka: `temp3` and `temp4` are defined as fixed-size `char` arrays earlier in the function. So yes, unless someone changes their size to `0`, we are sure ;)

source/base/fileinputoutput.cpp
LeForgeron: Yes, fine since C99/Posix clarification.
I'm too old, and used to be burnt by vsnprintf in old libraries.

PR #354, opened 2018-04-08
adding spiral render pattern (both direction)

LeForgeron:
add render pattern 6 & 7:
- 6: spiral from center to top left
- 7: spiral from top left to center

It was reported as a missing feature in the newsgroups.


LeForgeron:
here a screenshot of render window with +RP6

spiral-bench

LeForgeron:
I changed the destination to master as release/v3.8.0 seems stall now.

PR #355, opened 2018-04-08
fix the orientation of new uv-mapping of ovus to be similar to cylinder

LeForgeron:
and lemon.

Start from a commit which does not fail the automatic verification


LeForgeron:
This is how it should look:

[demo.zip](https://github.com/POV-Ray/povray/files/1887825/demo.zip)


refimguvmapping

PR #361, opened 2018-06-10
Leaving mesh inside_vector untouched on transforms.

wfpokorny:
Formally, both the ray origin and the direction derived from the
inside_vector for the inside mesh test would see an inverse transform
into the mesh's coordinate space.  This would sometimes result in a
change in the inside_vector/ray direction with respect to the mesh.
Such direction changes are problematic given test ray directions roughly
perpendicular to a triangle's normal are noisy with respect to
ray-triangle intersections.

Running the attached test scene we correctly get the attached image where a planes texture is controlled by an object pattern based upon the mesh. The star mesh has a flat back very set just inside the plane surface. If the scene is run against the current master, the inside of the star mesh is missed. 

meshinisidevect
[MeshInisideVect.pov.txt](https://github.com/POV-Ray/povray/files/2088042/MeshInisideVect.pov.txt)

### DIFF
--- a/source/core/shape/mesh.cpp
+++ b/source/core/shape/mesh.cpp
@@ -10,7 +10,7 @@
 /// @parblock
 ///
 /// Persistence of Vision Ray Tracer ('POV-Ray') version 3.8.
-/// Copyright 1991-2017 Persistence of Vision Raytracer Pty. Ltd.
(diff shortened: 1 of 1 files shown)


wfpokorny:
Branch at closing still exists and is occasionally re-based to master.

PR #368, opened 2019-03-04
do not use twice the same id label

fchapoton:
fix suggested by lgtm


c-lipka:
This change would also necessitate changing the stylesheet.

Please note that a lot of the doxygen stuff is Doxygen defaults plus/minus some minor modifications. If the isseue is also present in the Doxygen 1.8.7 defaults, and has been fixed by now, please amend the files in just the same way as the Doxygen defaults have been amended. If it hasn't been fixed, please leave it as-is and file a bug report to the Doxygen project.

PR #393, opened 2020-02-05
Pull request

fatman2021:
Updated the latest source code and included executables fo Amiga as well as the source code and executable for the latest version for MS-DOS.


c-lipka:
This is not a source code archive of old POV-Ray versions. While the last Amiga and MS-DOS versions may indeed be of nostalgic interest to the POV-Ray community, this is not the place to collect them,

Issue #395, opened 2020-02-16
Cahce large Image Maps when rendering animations


mlsomers:
### Summary

Parsing takes allot of time when loading huge image maps (height and color maps for planets for example). To speed up animation rendering these images could be cached for reuse.

### Suggested Solution

I did an attempt of getting my feet wet in C++, not successful so far :-(

Here's some C# pseudo code of what I would like to do:

```
namespace pseudocode
{
  public class Parser // this would be pov_parser::Parser
  {
    public class Image { } // placeholder representing pov_base::Image


    // -- parser.h

    private struct ImageFileInfo // The Dictionary Value type
    {
      public DateTime LastModifiedDate;
      public Image Image;
    }

    private static Dictionary<string, ImageFileInfo> Cache = new Dictionary<string, ImageFileInfo>(); // static cache

    
    // -- parser.cpp

    // Replace the original Read_Image with a check if it is in the cache
    public Image Read_Image(int filetype, string filename, int options)
    {
      ImageFileInfo imageInfo;
      if (Cache.TryGetValue(filename, out imageInfo))
      {
        if(File.GetLastWriteTime(filename) == imageInfo.LastModifiedDate)
        return imageInfo.Image;
      }

      imageInfo = new ImageFileInfo { 
        LastModifiedDate = File.GetLastWriteTime(filename),
        Image = Original_Read_Image(filetype, filename, options)
      };

      Cache.Add(filename, imageInfo);
      return imageInfo.Image;
    }

    // Placeholder for the original Read_Image method...
    private Image Original_Read_Image(int filetype, string filename, int options) {
      return null;
    }

    // Allow UI to clear memory
    public void ClaerCache()
    {
      Cache.Clear();
    }
  }
}
```

This is my lousy attempt at C++ (for now omitting the LastWriteTime check)

parser.h:
```
+      static std::map<pov_base::UCS2, pov_base::Image> ImageCache;

       Image *Read_Image(int filetype, const UCS2 *filename, const ImageReadOptions& options);
+      Image *Original_Read_Image(int filetype, const UCS2* filename, const ImageReadOptions& options);
```

parser.cpp:

```
Image *Parser::Read_Image(int filetype, const UCS2* filename, const ImageReadOptions& options)
{
    std::map<UCS2, Image>::iterator idx = ImageCache.find(*filename);
    if (idx != ImageCache.end()) {
        Image& ret = ImageCache[*filename];
        return &ret;
    }

[...]

PR #396, opened 2020-02-22
Cache Images for faster parsing subsequent images (or frames in animation)

mlsomers:
Loading large images during parsing can take ages. When rendering animations these images are reloaded before rendering each frame.

With this solution the images are only loaded once and cached for reuse when parsing the next frame. This can dramatically decrease parsing time.

Be free to point out any errors, this is my first serious C++ code written ever (normally I do C#).
The code has been tested and I don't see any significant memory leakage, in fact I am already using this in production (and it is saving lots of resources!).

I asked for this feature in issue #395 , instead of waiting for it I embarked on this rewarding quest.

### DIFF
--- a/source/core/support/imageutil.cpp
+++ b/source/core/support/imageutil.cpp
@@ -1387,10 +1387,14 @@ ImageData *Copy_Image(ImageData *Old)
 
 void Destroy_Image(ImageData *image)
 {
-    if ((image == nullptr) || (--(image->References) > 0))
-        return;
+    return;
 
-    delete image;
+    // Images are now cached
+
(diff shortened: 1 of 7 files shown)


mlsomers:
Finally passes all checks, but there is one TODO:
Somehow I missed the ImageData wrapper around Image. The wrapper should be deleted but not the child Image... These wrappers are currently leaking (in my case insignificant, but I'll still try to solve this).

mlsomers:
Since I cannot isolate these changes from other pull requests from my master branch, I'll open a new pull request from a different branch.

PR #397, opened 2020-02-23
Speed up loading uncompressed bmp files

mlsomers:
Improved the speed of loading standard 24bit BMP files (and 32bit with alpha).

Instead of reading one byte at a time, reading a whole line at a time reduces allot of "context switching".
On my (slow) machine it reduced loading time from 8 minutes down to 2 minutes! (my machine is heavily swapping memory while doing the test).


mlsomers:
yikes! why is this not separated from my previous pull request?
It should only include the file bmp.cpp. The rest is for caching!

mlsomers:
Reverted all changes related to the other pull request.
Also moved conditional alpha outside the nested loops (gain was not significant).

PR #422, opened 2021-07-02
Changes to support OpenEXR 3.0.

TonVanDenBroek:
Changes to support OpenEXR 3,0

### DIFF
--- a/source/base/image/openexr.cpp
+++ b/source/base/image/openexr.cpp
@@ -46,21 +46,11 @@
 
 #ifndef OPENEXR_MISSING
 
-#ifdef NON_UNIX_OPENEXR_HEADERS
-
 #include <ImfRgbaFile.h>
 #include <ImfStringAttribute.h>
 #include <ImfMatrixAttribute.h>
 #include <ImfArray.h>
 
-#else
-
-#include <OpenEXR/ImfRgbaFile.h>
-#include <OpenEXR/ImfStringAttribute.h>
-#include <OpenEXR/ImfMatrixAttribute.h>
-#include <OpenEXR/ImfArray.h>
-
-#endif
 #include "metadata.h"
 
 // this must be the last file included
@@ -105,7 +95,7 @@ class POV_EXR_OStream : public Imf::OStream
 				throw POV_EXCEPTION(kFileDataErr, "Error while writing EXR output");
 		}
 
-		Int64 tellp()
+		unsigned long tellp()
 		{
 			unsigned long pos = os.tellg();
 			if((int) pos == -1)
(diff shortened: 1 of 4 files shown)


source/base/image/openexr.cpp
```diff
@@ -105,17 +95,17 @@ class POV_EXR_OStream : public Imf::OStream
 				throw POV_EXCEPTION(kFileDataErr, "Error while writing EXR output");
 		}
 
-		Int64 tellp()
+		unsigned long tellp()
 		{
 			unsigned long pos = os.tellg();
 			if((int) pos == -1)
```
c-lipka, line 101: While you're at it, all `unsigned long` in the file should be considered a bug. They should either be `POV_LONG` (v3.7) / `POV_OFF_T` (v3.8) or whatever we'll be using instead of `Int64`.

Same goes for the comparison with -1 here and further below: There's a bug lurking in there as well, and it should instead be something like `if(pos == POV_LONG(-1))` (or correspondingly for v3.8).

unix/config/ax_check_openexr.m4
```diff
@@ -46,17 +46,11 @@ AC_DEFUN([AX_CHECK_OPENEXR],
...
-      AX_COMPARE_VERSION([$ax_check_openexr_version], [ge], [1.4],
-        [LIBS="$ax_check_openexr_libs -lIlmThread $LIBS"],
-	[LIBS="$ax_check_openexr_libs $LIBS"]
-      )
-
-      # check include file
       AC_CHECK_HEADER(
         [OpenEXR/ImfCRgbaFile.h],
-        [AC_CHECK_LIB([IlmImf], [ImfInputReadPixels], [], [ax_check_openexr="not found"])],
+        [ax_check_openexr="ok"],
```
c-lipka, line 53: That looks too easy... but yes - _(facepalm)_ - we're already probing for the libs via `pkg-config` anyway, so why do it again the hard and fragile way.

source/base/image/openexr.cpp
```diff
@@ -46,21 +46,11 @@
...
 #include <ImfArray.h>
 
-#else
-
-#include <OpenEXR/ImfRgbaFile.h>
-#include <OpenEXR/ImfStringAttribute.h>
-#include <OpenEXR/ImfMatrixAttribute.h>
-#include <OpenEXR/ImfArray.h>
-
-#endif
```
c-lipka, line 63: Given that removing this seems to work fine, I'm stumped by the question, _how did that not cause any issues before?_

If you have the time, could you perhaps dig into the `windows/vs10/povbase.vcxproj` XML file, and remove any reference to `NON_UNIX_OPENEXR_HEADERS`?

c-lipka (changes requested):
Looks solid, except for the issues with the Windows version. If you could address those (and maybe also fix the other type-related bugs along the way), that would be great.

c-lipka (approved):

PR #423, opened 2021-07-02
Changes to support OpenEXR 3.0.

TonVanDenBroek:
Changes to support OpenEXR 3,0


c-lipka (changes requested):
See review of #422: Same considerations apply, except that instead of `POV_LONG` we'll need `POV_OFF_T`, and we don't have to bother with any of the `NON_UNIX_OPENEXR_HEADERS` related stuff.

c-lipka (approved):

